PR #36, opened 2025-08-16
Fixes #33 (using bleach to sanitize the text filds)

### LINKED ISSUE #33: Sanitize description text in database models
As a platform developer, I want to sanitize description text bodies in my database models, so that I can prevent javascript and xss attacks.

This applies to all the "description" fields in the Organization, EventSeries and Event models.

A library such as markdown fields might be used.

Dawdaborje:
I used bleach to sanitize the text fields both on save and create


src/events/models/event_series.py
```diff
@@ -58,6 +58,11 @@ def for_user(self, user: RevelUser | AnonymousUser) -> EventSeriesQuerySet:
         """Get the queryset based on the user."""
         return self.get_queryset().for_user(user)
 
+    def create(self, **kwargs: t.Any):  # noqa: D102
+        if "description" in kwargs:
+            kwargs["description"] = bleach.clean(kwargs["description"])
+        return super().create(**kwargs)
+
```
biagiodistefano, line 65: Overriding create and and save methods is not always a good idea. It might be best to add this step to the clean() method instead. It will be called automatically on save() for all the `TimeStampedModel`.

src/events/models/event_series.py
```diff
@@ -72,3 +77,8 @@ class Meta:
             models.UniqueConstraint(fields=["organization", "name"], name="unique_event_series_name"),
             models.UniqueConstraint(fields=["organization", "slug"], name="unique_event_series_slug"),
         ]
+
+    def save(self, *args: t.Any, **kwargs: t.Any) -> None:  # noqa: D102
+        if self.description:
+            self.description = bleach.clean(self.description)
+        return super().save(*args, **kwargs)
```
biagiodistefano, line 84: Same as above

src/events/models/organization.py
```diff
@@ -84,6 +85,11 @@ def for_user(self, user: RevelUser | AnonymousUser, allowed_ids: list[UUID] | No
         """Get queryset for user."""
         return self.get_queryset().for_user(user)
 
+    def create(self, **kwargs: t.Any):  # noqa: D102
+        if "description" in kwargs:
+            kwargs["description"] = bleach.clean(kwargs["description"])
+        return super().create(**kwargs)
+
```
biagiodistefano, line 92: Same as above

src/events/models/organization.py
```diff
@@ -144,6 +150,11 @@ def is_stripe_connected(self) -> bool:
         """Check if the organization has a Stripe account connected."""
         return self.stripe_account_id is not None and self.stripe_charges_enabled and self.stripe_details_submitted
 
+    def save(self, *args: t.Any, **kwargs: t.Any) -> None:  # noqa: D102
+        if self.description:
+            self.description = bleach.clean(self.description)
+        return super().save(*args, **kwargs)
+
```
biagiodistefano, line 157: Same as above

Dawdaborje:
> I think a better approach would be to use an existing library. If not viable (a bit of research is needed), then override/extend the `clean()` methods of the models instead.
> 
> Another approach might be to use django-bleach or django-markdown-fields.
> 
> In any case, make sure to run `make check` and `make test` locally.

So in case of django-bleach, You prefer to use the model field??

e.g:

```python
from django import models
from django_bleach.models import BleachField

class Post(models.Model):

    title = models.CharField()
    content = BleachField()
```

biagiodistefano:
> > I think a better approach would be to use an existing library. If not viable (a bit of research is needed), then override/extend the `clean()` methods of the models instead.
> > Another approach might be to use django-bleach or django-markdown-fields.
> > In any case, make sure to run `make check` and `make test` locally.
> 
> So in case of django-bleach, You prefer to use the model field??
> 
> e.g:
> 
> ```python
> from django import models
> from django_bleach.models import BleachField
> 
> class Post(models.Model):
> 
>     title = models.CharField()
>     content = BleachField()
> ```

@Dawdaborje correct!

PR #127, opened 2025-12-02
Disable grafana

offby1:
Without this, the compose project fails to start, since grafana depends on pyroscope, which is itself disabled.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -87,7 +87,7 @@ jobs:
 
       - name: Run unit tests
         run: |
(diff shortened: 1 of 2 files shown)


biagiodistefano:
I think it would be better to rename `docker-compose-dev.yaml` to `docker-compose-observability.yaml` and fix it properly, in local dev is better to use `docker-compose-ci.yaml` (used in the CI) with observability disabled in the env vars. And `docker-compose-ci.yaml` should be renamed `docker-compose-dev.yaml`. [The docker-compose with the full working stack is in the infra repo](https://github.com/letsrevel/infra/blob/main/docker-compose.yml). Didn't have time to fix the docs about it. If you are willing to do this I'll approve the PR and merge. Thanks for the contribution!

PR #130, opened 2025-12-02
Rename and reorganize compose files.

offby1:
This allows "make setup" to complete.

This replaces <https://github.com/letsrevel/revel-backend/pull/127>.


biagiodistefano:
This is great, thank you. The only problem is that the CI will fail because the test.yaml still mentions the -ci.yaml

PR #373, opened 2026-04-17
fix(blacklist): reset stale BANNED membership on unban

biagiodistefano:
## Summary

- Unbanning a user no longer leaves them invisibly banned from the organization. When a `Blacklist` row was deleted, the `OrganizationMember(status=BANNED)` that `apply_blacklist_consequences` had created/updated was left in place, and `Organization.for_user()` kept excluding the org (and its events) for that user regardless of blacklist state.
- `remove_from_blacklist` now transitions the stale `BANNED` membership to `CANCELLED` (no implicit re-grant of prior `ACTIVE`/staff privileges). The fix is guarded by `check_user_hard_blacklisted`, so if another blacklist entry still covers the user in that org, the ban persists.
- Bumped to **1.48.1**.

## Root cause

`apply_blacklist_consequences` uses `OrganizationMember.objects.update_or_create(..., defaults={"status": BANNED})`, triggered by a `post_save` signal on `Blacklist`. There was no `post_delete` counterpart, and `remove_from_blacklist` only called `entry.delete()`. `Organization.for_user()` excludes orgs for which the user has a `BANNED` membership, independent of the `Blacklist` row — so the ban effectively persisted forever.

## Design choice

- **`CANCELLED`, not `ACTIVE`**: the ban path doesn't record what the status was pre-ban. Restoring to `ACTIVE` would implicitly re-grant membership (and any prior privileges) to a user who might never have had one. `CANCELLED` is treated as "no membership" by the visibility queryset — safe default.
- **Guard against multiple entries**: `link_blacklist_entries_for_user` can attach additional unlinked `Blacklist` rows to a user via `.update()` (bypassing `add_to_blacklist`'s dedup). Removing one must not unban if others still cover the user.
- **Service-layer fix, not a `post_delete` signal**: explicit and traceable, consistent with the existing `add_to_blacklist` / `apply_blacklist_consequences` surface.

## Test plan

- [x] `test_remove_from_blacklist_resets_banned_membership` — PUBLIC org, linked user: ban → assert invisible via `for_user`, unban → assert `CANCELLED` and visible again.
- [x] `test_remove_from_blacklist_keeps_banned_if_other_entries_exist` — second email-only entry keeps the ban in place after the linked entry is removed.
- [x] `test_remove_from_blacklist_without_user_is_noop` — unlinked entry deletion doesn't create or touch memberships.
- [x] `make check` passed locally
- [x] `make test` passed locally

## Bug Fixes
* Fixed unbanning behavior: when users are unbanned from organizations, stale membership records are now cleared and membership status is updated appropriately. Prior privileges are prevented from being implicitly re-granted, unless other blacklist entries remain in effect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Decomposed long functions into focused helpers for readability
 - Extracted `is_owner_or_staff` and `has_org_permission` permission helpers
 
+## [1.48.1] - 2026-04-17
+
 ### Fixed
 - Unbanning a user now clears the stale `BANNED` `OrganizationMember` row that the ban created. The membership transitions to `CANCELLED` (not `ACTIVE`, to avoid implicitly re-granting prior membership privileges), so the org and its events become visible again. If another blacklist entry still covers the user in the org, the `BANNED` status is preserved.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Decomposed long functions into focused helpers for readability
 - Extracted `is_owner_or_staff` and `has_org_permission` permission helpers
 
+### Fixed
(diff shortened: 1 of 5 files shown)


biagiodistefano:
### Code review

Issue #382, opened 2026-04-25
feat(events): expose cancellation policy on public TicketTierSchema for checkout transparency


biagiodistefano:
## Summary

The public `TicketTierSchema` (returned by `/api/events/{event_id}/ticket-tiers`, surfaced to buyers in the FE checkout flow) doesn't include cancellation/refund policy fields. As a result, buyers can't see the cancellation policy *before* committing to purchase — they'd have to buy first and then call `/cancellation-preview` to find out the terms.

## What's missing

These fields exist on `TicketTierDetailSchema` (admin) but not on `TicketTierSchema` (public):

- `allow_user_cancellation: bool`
- `cancellation_deadline_hours: int | None`
- `refund_policy: RefundPolicy | None`

## Why it matters

The FE wants to render the refund schedule directly inside the purchase confirmation dialog so buyers know what they're agreeing to (mirroring Eventbrite / Ticketmaster behavior). Reproducing this without server-supplied policy data isn't possible — the brackets and flat fee are the contract, and showing them post-purchase is too late.

## Suggested change

Add the three fields to `TicketTierSchema` (read-only at this layer — no server-side validation needed since the public schema is response-only). Public ticket tier listings should include:

```python
class TicketTierSchema(...):
    ...
    allow_user_cancellation: bool = False
    cancellation_deadline_hours: int | None = None
    refund_policy: RefundPolicy | None = None
```

Once shipped, the FE will:
- Show a "Free cancellation until …" or "Refund schedule" panel inside the ticket confirmation dialog.
- Gate the "Cancel ticket" affordance in `MyTicketModal` on `tier.allow_user_cancellation` directly (currently we just call `/cancellation-preview` and surface its `not_permitted` reason — works, but a wasted round-trip).

## Frontend tracking

`letsrevel/revel-frontend` branch `feature/379-user-ticket-cancellation` is shipping checkout-side policy display behind a defensive check that no-ops until these fields are present, so this issue can land independently.

PR #383, opened 2026-04-25
fix(events): refund_policy 400, Stripe Connect refund 502, and TICKET_CANCELLED staff copy

### LINKED ISSUE #381: fix(events): refund_policy POST/PUT returns 400 "Value must be valid JSON."
## Summary

Creating or updating a `TicketTier` with a `refund_policy` payload returns 400 with `{"errors": {"refund_policy": ["Value must be valid JSON."]}}` even though the request body matches the OpenAPI schema.

Repro is trivial: `POST /api/event-admin/{event_id}/ticket-tier` with the body the OpenAPI spec advertises:

```json
{
  "name": "GA",
  "payment_method": "online",
  "price": "20",
  "currency": "EUR",
  "allow_user_cancellation": true,
  "refund_policy": {
    "tiers": [{ "hours_before_event": 24, "refund_percentage": 100 }],
    "flat_fee": "0"
  }
}
```

→ 400, `Value must be valid JSON.`

## Root cause

`refund_policy` is a `models.JSONField`. The controller hands the payload to the service via:

```python
payload_dict = payload.model_dump(exclude_unset=True)
```

`model_dump()` in default mode returns nested Pydantic models as `RefundPolicy` instances (or whatever the inner type is) rather than plain `dict`/`list`. The service then does:

```python
setattr(tier, field, value)  # value = RefundPolicy(...)
tier.save(update_fields=...)
```

`TimeStampedModel.save()` calls `full_clean()`, Django's `JSONField.validate()` runs `to_python()` on the assigned value, and because it's not a plain dict/list it falls through to `json.loads(str(value))` which raises `_('Value must be valid JSON.')`.

So the request body is fine — the backend never serializes the nested Pydantic model into a JSON-compatible primitive before handing it to Django.

Files involved:
- `src/events/controllers/event_admin/tickets.py` — `create_ticket_tier`, `update_ticket_tier`
- `src/events/service/ticket_service.py` — `create_ticket_tier`, `update_ticket_tier`
- `src/events/models/ticket.py:357` — `refund_policy = models.JSONField(...)`

## Suggested fix

Either:

1. Switch the dump to JSON mode in both controllers:
   ```python
   payload_dict = payload.model_dump(exclude_unset=True, mode="json")
   ```
   This recursively coerces nested Pydantic models (and `Decimal`, `datetime`, etc.) to JSON-compatible primitives.

2. Or explicitly normalize the field at the service boundary:
   ```python
   if "refund_policy" in tier_data and tier_data["refund_policy"] is not None:
       tier_data["refund_policy"] = tier_data["refund_policy"].model_dump(mode="json")
   ```

Option 1 is the simpler and broader fix — anything else nested (`refund_policy_snapshot` etc.) gets the same treatment for free, and it matches what the JSONField expects to receive from external callers.

## Tests

The bug isn't caught by existing service-layer tests because they pass dicts directly:

```python
# src/events/tests/test_service/test_ticket_tier_cancellation_fields.py:23
"refund_policy": {
    "tiers": [...],
    ...
}
```

Add a controller-layer test that posts the JSON body via the test client (not via the service) so the controller's `model_dump()` path is exercised.

## Frontend impact

The frontend (PR-equivalent branch `feature/379-user-ticket-cancellation` on `letsrevel/revel-frontend`) is sending the spec-compliant payload and gets blocked here. No FE changes needed once this is fixed.

biagiodistefano:
## Summary

Bundle of fixes uncovered while testing the user-initiated cancellation flow. Closes #381 and #382.

- **#381 — Admin tier 400 "Value must be valid JSON."**: switch `create_ticket_tier`/`update_ticket_tier` controllers to `model_dump(mode="json")` so nested `RefundPolicy` and `Decimal` leaves reach the `JSONField` as JSON-serializable primitives. Default `json.JSONEncoder` can't handle `Decimal`, so `full_clean()` was raising during persist.
- **Stripe Connect refund 502**: `_issue_stripe_refund` (cancellation_service.py) now passes `stripe_account=org.stripe_account_id` for direct charges, mirroring the platform-account exemption used everywhere else in `stripe_service.py`. Without it Stripe looked for the payment intent on the platform account, raised `InvalidRequestError`, and the cancel endpoint returned 502.
- **#382 — Public `TicketTierSchema`**: add `allow_user_cancellation`, `cancellation_deadline_hours`, and `refund_policy` so buyers see the cancellation terms before committing to a purchase.
- **TICKET_CANCELLED notifications**: organizers were receiving the holder-addressed copy ("You cancelled your ticket for X") because all four templates branched only on `cancellation_source`. Branch first on `context.ticket_holder_name` (the existing staff-vs-holder signal used by `ticket_created`/`ticket_updated`) and render third-person copy for staff.

## Test plan

- [ ] `make check` passes (note: `make i18n-check` will fail until `make makemessages && make compilemessages` is run — new `blocktranslate` strings added in the cancelled templates)
- [ ] `make test` passes — new tests:
  - admin: POST/PUT `/ticket-tier` accepts a nested `refund_policy` end-to-end
  - service: `_issue_stripe_refund` includes `stripe_account` when org has a Connect account, omits it when org uses the platform account
  - schema: public `tier_list` exposes the three cancellation fields and degrades cleanly when no policy is configured
  - templates: staff audience renders third-person copy across all four channels for every `cancellation_source`
- [ ] Manual: re-run the cancel curl from the report — should succeed (or fall back to the documented 409/refund codes), no more 502
- [ ] Manual: organizer/staff in-app + email + telegram for a user-initiated cancellation no longer say "You cancelled..."


biagiodistefano:
### Code review

Found 1 issue:

1. `_issue_stripe_refund` accesses `ticket.event.organization.stripe_account_id`, but it is called with `locked_ticket` — which only has `select_related("tier")`. Each online cancellation now triggers two lazy queries (`event` then `organization`) inside the `select_for_update()` + `transaction.atomic()` block, enlarging the lock window during the synchronous Stripe API call. The original `ticket` arg passed into `cancel_ticket_by_user` is loaded via `.full()` and would have these relations preloaded, but the locked re-fetch loses them. Add `event__organization` to the `select_related` on the locked re-fetch.

https://github.com/letsrevel/revel-backend/blob/4cab632509c8c549fe6aeadb1130b375ae2ee45d/src/events/service/cancellation_service.py#L361-L377

https://github.com/letsrevel/revel-backend/blob/4cab632509c8c549fe6aeadb1130b375ae2ee45d/src/events/service/cancellation_service.py#L438-L451

Generated with [Claude Code](https://claude.ai/code)

PR #409, opened 2026-05-15
fix(events): add admin endpoint to fetch series template event

biagiodistefano:
## Summary

- Adds `GET /organization-admin/{slug}/event-series/{series_id}/template-event` returning the full `EventDetailSchema` for the series template event, gated by `OrganizationPermission("edit_event_series")`.
- Fixes a hotfix-grade 404 on the FE TemplateEditDialog: it refetched the template via `GET /events/{event_id}`, but `Event.objects.for_user()` filters `is_template=False`, so admins editing a series template could not load the form.

## Why a new endpoint

The public path's template filter is correct — templates are admin-only artifacts and should not be reachable through the public detail route. Adding an admin path bypasses the filter behind the existing `edit_event_series` permission rather than weakening the public surface or routing template edits through a separate schema.

The frontend will be updated in a follow-up to consume the new endpoint after `pnpm generate:api`.

## Test plan

- [x] `make check` (ruff format/lint, mypy --strict, migration/i18n/file-length checks) — green
- [x] `make test` (full suite, 4469 passed, 15 skipped) — green
- [x] New `TestGetSeriesTemplateEvent` covers: owner gets full template detail (incl. `is_template=True`), nonmember → 404, missing series → 404, cross-org series id → 404, empty series (no template) → 404
- [ ] Manually verify FE template-edit dialog stops 404'ing after the FE patch lands

### DIFF
--- a/src/events/controllers/organization_admin/recurring_events.py
+++ b/src/events/controllers/organization_admin/recurring_events.py
@@ -180,6 +180,27 @@ class OrganizationAdminRecurringEventsController(OrganizationAdminBaseController
         series = self._get_series(slug, series_id)
         return schema.EventSeriesDriftSchema(stale_occurrences=recurrence_service.detect_cadence_drift(series))
 
+    @route.get(
+        "/event-series/{series_id}/template-event",
+        url_name="get_series_template_event",
+        response=schema.EventDetailSchema,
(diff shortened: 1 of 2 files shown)


biagiodistefano:
### Code review

PR #422, opened 2026-05-18
feat(accounts): self-served email change via JWT-confirmation flow

### LINKED ISSUE #421: feat(accounts): self-served email change via JWT-confirmation flow
## Summary

Add a self-served email change flow that lets an authenticated user change their account email by proving control of the **new** address. Mirrors the existing JWT-token email flows (verify-email, password-reset, delete-account) in `accounts/`.

## Motivation

Today, users cannot change their email address without operator intervention. Recent support requests have made this gap concrete (e.g. users locked out of their original mailbox asking us to swap it manually). A standard, self-served flow removes the manual touchpoint and matches what users expect from any modern account system.

## User flow

1. User is logged in and submits a new email + their current password.
2. They receive a confirmation link **at the new address**.
3. They click the link. The frontend POSTs the token.
4. Backend swaps `email` + `username`, blacklists every outstanding JWT, and returns a fresh token pair.
5. User is now signed in on the device that confirmed; **every other session is logged out** (this must be communicated up-front — see Frontend notes).
6. Both old and new addresses receive a confirmation email after the swap.

### `POST /account/email-change-request` (auth, `WriteThrottle`)

Request body:
```json
{ "new_email": "new@example.com", "password": "current-password" }
```

Behavior:
- Validates the user's current password (defense-in-depth, mirrors `delete_account_request`).
- Rejects if `new_email` is taken (case-insensitive match against `username`), is globally banned, or equals the current email.
- Rejects for Google SSO users (consistent with `request_password_reset` — their email is owned by Google).
- Issues a single-use JWT (new schema `EmailChangeJWTPayloadSchema`) with `user_id`, `new_email`, `current_email`, `exp`, `jti`, `type=email_change`. Reuses `settings.VERIFY_TOKEN_LIFETIME`.
- Dispatches two Celery tasks:
  - `send_email_change_confirmation` → **new** address, contains the confirmation link.
  - `send_email_change_notice` → **current** address, informational only ("a change to `n***@example.com` was requested; if it wasn't you, change your password now"). No cancel token in v1 — keep state in the JWT, not in a new model.

Response: `ResponseMessage` (success message, no enumeration leakage).

### `POST /account/email-change-confirm` (anon, `WriteThrottle`)

Request body:
```json
{ "token": "<jwt>" }
```

Behavior:
- Validates and blacklists the token (`token_to_payload` + `blacklist_token`).
- Re-checks `new_email` uniqueness at confirm time (race-safe — request and confirm can be minutes apart).
- Updates `user.email` and `user.username` atomically. Keeps `email_verified=True` (new address proven by clicking the link).
- **Blacklists all outstanding JWTs for the user.** Email is an identity primitive in this system, so we treat the change like a credential rotation. (Option A from brainstorming — full session invalidation.)
- Issues a fresh JWT token pair so the confirming device stays signed in.
- Dispatches `send_email_change_completed` to **both** old and new addresses.

Response:
```json
{
  "user": { "...RevelUserSchema..." },
  "token": { "access": "...", "refresh": "..." }
}
```

## Schemas

New in `accounts/schema.py`:

```python
class EmailChangeRequestSchema(Schema):
    new_email: EmailStr
    password: str

    @field_validator("new_email")
    @classmethod
    def lowercase_email(cls, v: str) -> str:
        """Normalize email to lowercase (mirrors RegisterUserSchema)."""
        return v.lower()

class EmailChangeConfirmSchema(Schema):
    token: str

class EmailChangeJWTPayloadSchema(_BaseEmailJWTPayloadSchema):
    new_email: EmailStr
    type: t.Literal["email_change"] = "email_change"

class EmailChangeResponseSchema(Schema):
    user: RevelUserSchema
    token: TokenSchema  # same shape returned by verify_email
```

**Username/email normalization** — store the lowercased form, matching registration:
- `EmailChangeRequestSchema.new_email` is lowercased at the schema layer (validator above), same approach as `RegisterUserSchema.lowercase_email`. The lowercased value is what gets baked into the JWT and what eventually lands in `user.email` / `user.username`.
- Do **not** apply `normalize_email_for_matching` from `accounts/utils/email_normalization.py`. That helper is intentionally lossy (strips `+tag`, drops dots for Gmail) and is reserved for global-ban matching, never for the stored value.
- The uniqueness check uses `username__iexact=new_email` to be defensive against any legacy non-lowercased rows.

Service layer additions in `accounts/service/account.py`:

- `create_email_change_token(user, new_email) -> str`
- `request_email_change(user, new_email, password) -> str`
- `confirm_email_change(token) -> RevelUser`

New constants and Celery tasks in `accounts/tasks.py`:

- `TOKEN_TYPE_EMAIL_CHANGE`
- `send_email_change_confirmation(new_email, token)` → link to new address
- `send_email_change_notice(current_email, new_email_masked)` → notice to old address
- `send_email_change_completed(old_email, new_email)` → fan-out to both after confirm

Email templates under `accounts/templates/accounts/email/` following the existing structure.

## Security considerations

- **Token sent to new address**, not current — that's the proof-of-control mechanism.
- **Password required at request time** — limits abuse if a session is briefly hijacked.
- **Session-wide blacklist on confirm** — if an attacker requested the change and somehow obtained the token, the legitimate user is at least logged out of any compromised devices.
- **Notice to old address** — gives the legitimate user a signal even if they didn't initiate.
- **Single-use token** via existing blacklist mechanism.
- **Race-safe uniqueness check** on confirm to prevent two users grabbing the same email between request and confirm.
- **No enumeration**: the request endpoint should return a consistent success message even if the new address is taken (or surface a generic error — to be decided during implementation; lean toward generic-error since the user already knows the address).
- Rate-limit with `WriteThrottle` on both endpoints; consider tightening if abuse is observed.

## Edge cases

- New email == current email → 400 with clear message (not an enumeration concern).
- New email already in use by another account → reject. The user knows what they typed, so enumeration risk is minimal.
- User changes mind → token expires naturally; no explicit cancel in v1.
- User requests twice → second token supersedes the first only at use time (both remain valid until used or expired). Accept this for v1; if it becomes a problem, blacklist outstanding email-change tokens for the user at request time.
- Globally banned new email → silently no-op (consistent with verification flow).

## Frontend notes (`revel-frontend`)

- New settings UI page or section: "Change email address". Fields: new email, current password. Submit calls `email-change-request`.
- **Show a clear warning before submitting**: "You will be logged out of every other device when you confirm the change." Block submission behind an acknowledgement (checkbox or confirm dialog), not a passive footnote.
- Confirmation page reads the token from the URL and POSTs to `email-change-confirm`. On success, replace the in-memory access token with the new one returned by the backend (so the current tab stays signed in seamlessly).
- Existing translations: add strings for the new email subjects, body templates, and the FE warning.
- Regenerate the API client (`pnpm generate:api`) once backend ships.

## Testing

- Service-layer tests for `request_email_change` (success, wrong password, duplicate email, SSO user, banned email, same-email).
- Service-layer tests for `confirm_email_change` (success, expired token, blacklisted token, race on email uniqueness, JWT blacklist verification).
- Controller tests for both endpoints, including the email-fanout to old/new addresses (mock Celery tasks).
- Integration test: full request → confirm → re-login on a stale token fails (proves session invalidation).

## Out of scope (deferred)

- Cancel-from-old-address link (would require a second JWT and a small UI surface; YAGNI for v1).
- Admin-side email change tool.
- Multi-step verification (e.g. confirm at both old and new) — overkill for the common case; the password gate + notice covers most threats.

biagiodistefano:
Closes #421.

## Summary
- Adds `POST /account/email-change-request` (auth, WriteThrottle) and `POST /account/email-change-confirm` (anon, WriteThrottle), mirroring the existing JWT-token email flows (verify-email, password-reset, delete-account).
- Confirmation link is sent to the **new** address (proof of control). The current address receives a separate notice with a masked rendering of the new address.
- On confirm: validate + blacklist the single-use token, swap `email` + `username`, blacklist every outstanding JWT for the user (email is an identity primitive), return a fresh token pair so the confirming device stays signed in. Old and new addresses both receive a completion email.
- Promotes `_blacklist_user_tokens` out of `global_ban_service.py` into `accounts/jwt.py` (`blacklist_user_tokens`) so both call sites share it.

## Key files
- `src/accounts/schema.py` — `EmailChangeRequestSchema`, `EmailChangeConfirmSchema`, `EmailChangeJWTPayloadSchema`, `EmailChangeResponseSchema`
- `src/accounts/service/account.py` — `create_email_change_token`, `request_email_change`, `confirm_email_change`, `_mask_email`
- `src/accounts/tasks.py` — `send_email_change_confirmation`, `send_email_change_notice`, `send_email_change_completed_old`, `send_email_change_completed_new`
- `src/accounts/controllers/account.py` — two new endpoints
- `src/accounts/templates/accounts/emails/email_change_*.{txt,html}` — 12 new templates (4 emails × subject/txt/html)
- `src/common/testing.py` — `TOKEN_TYPE_EMAIL_CHANGE`

## Notes / decisions
- **Duplicate / banned target**: clear 400 on duplicate (user already knows the address); silent no-op on globally-banned target (mirrors verification flow, avoids signalling ban presence).
- **Same-email** and **Google-SSO** both rejected with a clear 400.
- **Race-safe**: uniqueness re-checked at confirm time; on race-loss, token is blacklisted (single-use) so it can't be retried later.
- **Two completed-emails tasks** rather than one fan-out task — isolated failure modes.

## Test plan
- [ ] `make check` (green locally)
- [ ] `make test` covering: wrong password, same email, duplicate email (case-insensitive), Google SSO rejection, globally-banned target no-op, expired token, single-use enforcement, race-safe uniqueness, full-flow blacklist of pre-existing refresh tokens
- [ ] Frontend follow-up: regenerate API client (`pnpm generate:api`) once this merges; add the settings UI with the "you will be signed out of every other device" warning

## Release Notes

**New Features**
- Users can now change their email address via self-service, requiring password verification.
- Email changes require confirmation through a secure link sent to the new address.
- All existing sessions are automatically signed out after confirming the new email for security.
- Confirmation and notification emails are sent to both old and new addresses.

[Review Change Stack](https://app.coderabbit.ai/change-stack/letsrevel/revel-backend/pull/422?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "revel"
-version = "1.55.0"
+version = "1.56.0"
 description = "The Revel Event Manager"
 readme = "README.md"
 license = "MIT"
--- a/src/accounts/controllers/account.py
+++ b/src/accounts/controllers/account.py
@@ -220,6 +220,47 @@ class AccountController(UserAwareController):
         return ResponseMessage(message=str(_("Password reset successfully.")))
 
     @route.post(
+        "/email-change-request",
+        tags=["Account"],
+        response=ResponseMessage,
+        url_name="email-change-request",
+        throttle=WriteThrottle(),
+    )
+    def email_change_request(self, payload: schema.EmailChangeRequestSchema) -> ResponseMessage:
+        """Begin a self-served email change.
+
+        Validates the current password, ensures the new address is available, issues a
+        single-use JWT, and emails the confirmation link to the **new** address. A
+        separate informational notice is sent to the current address. Returns a generic
+        success message on duplicate-banned-email targets to avoid signalling ban presence.
+        """
+        account_service.request_email_change(
+            user=self.user(),
+            new_email=payload.new_email,
+            password=payload.password,
+        )
+        return ResponseMessage(message=str(_("A confirmation link has been sent to the new email address.")))
+
+    @route.post(
+        "/email-change-confirm",
+        tags=["Account"],
+        response=schema.EmailChangeResponseSchema,
+        url_name="email-change-confirm",
+        auth=None,
+        throttle=WriteThrottle(),
+    )
+    def email_change_confirm(self, payload: schema.EmailChangeConfirmSchema) -> schema.EmailChangeResponseSchema:
+        """Confirm an email change using the token sent to the new address.
+
+        Swaps email/username, blacklists all existing JWTs for the user (email is an
+        identity primitive), and issues a fresh token pair so the confirming device
+        stays signed in. The confirmation token is single-use.
+        """
+        user = account_service.confirm_email_change(payload.token)
+        token = get_token_pair_for_user(user)
+        return schema.EmailChangeResponseSchema(user=RevelUserSchema.from_orm(user), token=token)
+
+    @route.post(
         "/me/upload-profile-picture",
         response=RevelUserSchema,
         url_name="upload-profile-picture",
--- a/src/accounts/controllers/account.py
+++ b/src/accounts/controllers/account.py
@@ -231,8 +231,16 @@ class AccountController(UserAwareController):
 
         Validates the current password, ensures the new address is available, issues a
         single-use JWT, and emails the confirmation link to the **new** address. A
-        separate informational notice is sent to the current address. Returns a generic
-        success message on duplicate-banned-email targets to avoid signalling ban presence.
+        separate informational notice is sent to the current address. Returns 400 if
+        the new address is already in use; returns the generic success message when the
+        target is globally banned, to avoid signalling ban presence.
+
+        Args:
+            payload: Schema with ``new_email`` (lowercased by the schema validator)
+                and the user's current ``password``.
+
+        Returns:
+            A generic success message indicating that the confirmation link was sent.
         """
         account_service.request_email_change(
             user=self.user(),
@@ -255,6 +263,12 @@ class AccountController(UserAwareController):
         Swaps email/username, blacklists all existing JWTs for the user (email is an
         identity primitive), and issues a fresh token pair so the confirming device
         stays signed in. The confirmation token is single-use.
+
+        Args:
+            payload: Schema containing the single-use email-change token.
+
+        Returns:
+            The updated user and a fresh JWT token pair for the confirming device.
         """
         user = account_service.confirm_email_change(payload.token)
         token = get_token_pair_for_user(user)
--- a/src/accounts/jwt.py
+++ b/src/accounts/jwt.py
@@ -126,6 +126,28 @@ def check_blacklist(jti: str) -> None:
         raise HttpError(401, "Token is blacklisted.")
 
 
+def blacklist_user_tokens(user: t.Any) -> int:
+    """Blacklist all outstanding JWT tokens for a user.
+
+    Used when an event invalidates every active session — e.g. global ban,
+    or an identity-primitive change like email rotation.
+
+    Args:
+        user: The user whose tokens should be blacklisted.
+
+    Returns:
+        Number of tokens blacklisted.
+    """
+    outstanding = OutstandingToken.objects.filter(user=user).exclude(blacklistedtoken__isnull=False)
+    count = 0
+    for token in outstanding:
+        BlacklistedToken.objects.get_or_create(token=token)
+        count += 1
+    if count:
+        logger.info("user_tokens_blacklisted", user_id=str(user.id), count=count)
+    return count
+
+
 @transaction.atomic
 def blacklist(
     token: str,
--- a/src/accounts/schema.py
+++ b/src/accounts/schema.py
@@ -214,6 +214,11 @@ class UnsubscribeJWTPayloadSchema(_BaseEmailJWTPayloadSchema):
     type: t.Literal["unsubscribe"] = "unsubscribe"
 
 
+class EmailChangeJWTPayloadSchema(_BaseEmailJWTPayloadSchema):
+    new_email: EmailStr
+    type: t.Literal["email_change"] = "email_change"
+
+
 class DeleteAccountSchema(Schema):
     password: str
 
@@ -261,6 +266,26 @@ class LanguageUpdateSchema(Schema):
 
 
 class VerifyEmailResponseSchema(Schema):
+    user: RevelUserSchema
+    token: TokenObtainPairOutputSchema
+
+
+class EmailChangeRequestSchema(Schema):
+    new_email: EmailStr
+    password: str
+
+    @field_validator("new_email")
+    @classmethod
+    def lowercase_email(cls, v: str) -> str:
+        """Normalize email to lowercase."""
+        return v.lower()
+
+
+class EmailChangeConfirmSchema(Schema):
+    token: str
+
+
+class EmailChangeResponseSchema(Schema):
     user: RevelUserSchema
     token: TokenObtainPairOutputSchema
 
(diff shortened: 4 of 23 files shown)


biagiodistefano:
### Code review — merged findings (self + Copilot + CodeRabbit)

CI red on shard 1 fixed in a9a4c04 (`GlobalBan` fixture had to pass `value`, not `normalized_value`).

Addressed in 91f2a22:

1. **SSO check ordered before password check** — SSO accounts were rejected with a misleading "Incorrect password" because the password check ran first. (Copilot)
   https://github.com/letsrevel/revel-backend/blob/91f2a22b25badfb4f3422ef7017c4e04706fda82/src/accounts/service/account.py#L368-L378

2. **Global-ban re-check at confirm time** — a ban added between request and confirm could still let the swap go through. Mirrors `verify_email`. (Copilot)
   https://github.com/letsrevel/revel-backend/blob/91f2a22b25badfb4f3422ef7017c4e04706fda82/src/accounts/service/account.py#L418-L426

3. **`IntegrityError` -> 400 on save** — if two concurrent confirms both pass the pre-check, the loser's save now surfaces as a clean 400 instead of a 500; the token is blacklisted outside the rolled-back transaction. (Copilot, CodeRabbit)
   https://github.com/letsrevel/revel-backend/blob/91f2a22b25badfb4f3422ef7017c4e04706fda82/src/accounts/service/account.py#L432-L453

4. **Controller docstring accuracy + Google-style `Args`/`Returns`** — the request-endpoint doc was wrong (it said duplicates get a generic success; actually only banned targets do). Added `Args`/`Returns` to both controller methods and the four new Celery tasks per CLAUDE.md. (Copilot, CodeRabbit)
   https://github.com/letsrevel/revel-backend/blob/91f2a22b25badfb4f3422ef7017c4e04706fda82/src/accounts/controllers/account.py#L227-L243

5. **Tests** (Copilot, CodeRabbit):
   - `test_duplicate_email_case_insensitive` now passes mixed-case input (was pre-lowercased).
   - `test_google_sso_user_rejected` no longer sets a local password — exercises the real sentinel-password path.
   - `test_blacklisted_token_rejected` / `test_race_email_taken` now assert blacklist by the specific JTI, not "any blacklisted token exists".
   - `test_full_flow_old_refresh_tokens_invalidated` now asserts the fresh refresh token returned by `/confirm` is **not** in the blacklist.
   - New `test_globally_banned_at_confirm_time_rejected` covers finding #2.

   https://github.com/letsrevel/revel-backend/blob/91f2a22b25badfb4f3422ef7017c4e04706fda82/src/accounts/tests/test_email_change.py#L95-L116

#### Acknowledged, not changed in this PR

- **Token survives an old-address password reset.** The notice tells the user to reset their password if the request was unauthorized, but the email-change token isn't bound to the password hash, so it remains valid. Mitigated by the 15-minute TTL (`VERIFY_TOKEN_LIFETIME`); a stronger fix would bind the JWT to the password hash and is a larger schema change — out of scope for v1.
- **Concurrent-confirm idempotency window.** Two confirms with the *same* token could in principle both pass `check_blacklist` before either commits. The outcome is idempotent (same final email, same fresh tokens) plus duplicate completion emails. Locking the user row would close it; accepting the rare race for v1.
- **PII in task logs.** CodeRabbit flagged logging the new/old emails as PII. Pre-existing convention across every account-email task (`send_password_reset_link`, `send_account_deletion_link`, etc.); changing it here only would be inconsistent. Tracked separately if we decide to redact.
- **`lang` attribute on email `<html>`.** Same pattern as every existing account email template; would address all templates in one pass rather than just the new ones.

`make check` green, 18/18 new tests pass locally.

PR #426, opened 2026-05-20
feat(events): advanced waitlist management (#2)

### LINKED ISSUE #2: User Story: Advanced Waitlist Management
#### Story #19: Organizer's Batch-Based Automated Waitlist

* **As an:** Organization Owner or Staff Member
* **I want to:** configure an advanced automated waitlist system with batch processing and lottery options
* **So that:** when spots become available, the system automatically notifies groups of people with configurable fairness and urgency settings

**Acceptance Criteria:**

1. **Waitlist Configuration Settings:** When creating or editing an `Event`, I can configure:
   - **Batch Size** (integer, default: 0): Number of users to notify at once
     - `0` = notify everyone simultaneously (open access mode)
     - `1-N` = notify N users at a time
   - **Time Window** (duration, range: 1 hour - 7 days, default: 24 hours): How long each batch has to claim spots before the next batch is notified
   - **Cutoff Date** (datetime, optional): After this date, all remaining waitlist members are notified simultaneously regardless of batch size
   - **Lottery Mode** (boolean, default: False): If enabled, batch members are randomly selected instead of FIFO

2. **Automatic Triggering:** The waitlist batch process is automatically triggered when spots open due to:
   - An attendee's `Ticket` being cancelled
   - An attendee's `EventRSVP` status changing from "Yes" to "No" or "Maybe"
   - The `max_attendees` of the `Event` being increased

3. **Batch Notification Flow:**
   - When spots become available, the first `batch_size` users (or all if batch_size=0) are selected from the waitlist
   - Selection follows FIFO order by default, or random selection if lottery mode is enabled
   - Selected users receive notifications with a clear time limit to claim their spots
   - During the time window, only users in the current batch can register (first-come-first-served within the batch)
   - If the time window expires and spots remain unclaimed, those offers expire and the next batch is notified
   - After the cutoff date (if set), all remaining waitlist members are notified simultaneously regardless of batch size

4. **Waitlist Offers as "Golden Tickets":**
   - Each notified user receives a temporary offer (stored in `WaitlistOffer` model)
   - Offers grant temporary eligibility to bypass "Event is Full" restrictions
   - Offers have explicit expiration times based on the time window
   - Multiple pending offers can exist simultaneously (equal to available spots)

5. **Offer Lifecycle Management:**
   - Offers can be in states: `PENDING`, `CLAIMED`, `EXPIRED`
   - Claimed offers convert to actual tickets/RSVPs and are removed from the waitlist
   - Expired offers trigger the next batch notification automatically
   - A scheduled task checks for expired offers periodically and processes the next batch

6. **Spot Reservation & Attendee Count Management:**
   - **CRITICAL:** When a ticket/RSVP is cancelled and waitlist offers are created, the spot is **reserved** for those offers
   - The `attendee_count` does **NOT** immediately decrease when a cancellation occurs if offers are created
   - Spots remain "held" while offers are pending to prevent double-booking
   - Only when all related offers expire without being claimed does the spot become truly available
   - This ensures fairness: the notified batch has exclusive access to the spot during their time window
   - If no offers are created (empty waitlist), then the spot is released immediately

#### Story #18: Attendee's Waitlist Experience with Batch Notifications

* **As a:** User on an event's waitlist
* **I want to:** be notified when it's my turn to claim a spot with a clear time limit
* **So that:** I understand when I have the opportunity to register and can act within the time window

**Acceptance Criteria:**

1. **Clear Batch Notification:** When I'm selected in a batch, I receive a notification via Email and/or Telegram containing:
   - The name of the event
   - A clear call to action (e.g., "You have been selected! Claim Your Spot Now!")
   - The exact date and time when my offer expires
   - Information about how many spots are available
   - A direct link to claim my spot

2. **Golden Ticket Access:** Clicking the link grants me temporary eligibility to register for the event, bypassing the "Event is Full" restriction during my time window

3. **Time-Limited Claim:** I must complete registration (RSVP or purchase a ticket) within the time limit or my offer expires

4. **Successful Claim:** If I complete registration within the time limit, my spot is confirmed and I am removed from the waitlist

5. **Expiration Behavior:** If I don't claim the spot in time:
   - My offer automatically expires
   - I may receive a notification that my offer has expired
   - The system automatically moves to the next batch

6. **Cutoff Date Behavior:** After the cutoff date, all waitlist rules are lifted and remaining waitlist members can claim spots freely (no more batching)

---

#### Step 0: Spot Reservation Architecture

**Core Principle: Spots are Reserved for Pending Offers**

When a cancellation occurs and waitlist offers are created, the cancelled spot is **not** released back to the general pool. Instead:

1. **Capacity Calculation:** Available spots = `max_attendees - (active_attendees + pending_offers)`
2. **Pending offers "hold" spots:** While offers are pending, the event remains at capacity from the perspective of non-waitlist users
3. **Only offer holders can claim:** During the time window, only users with active offers can register
4. **Release on expiration:** Only when all offers in a batch expire does the spot become available for the next batch

**Why This Matters:**

- **Prevents double-booking:** Non-waitlist users cannot "steal" spots reserved for waitlist batches
- **Fairness:** The notified batch gets exclusive access during their time window
- **Accurate capacity tracking:** The system always knows who has rights to claim spots
- **Race condition prevention:** Multiple users cannot compete for the same spot outside the batch system

**Implementation Impact:**

The `attendee_count` field behavior changes:
- **Before:** Decremented immediately on cancellation
- **After:** Only decremented when offers expire or no offers are created
- This requires modifying the `build_attendee_visibility_flags` task or signal handlers

**Critical: Update `EventManager._assert_capacity` Method**

The `_assert_capacity` method in `events/service/event_manager.py` (lines 564-595) currently only counts actual attendees (tickets/RSVPs). It **must** be updated to account for pending waitlist offers, which effectively reserve spots during their time window.

Without this update, non-waitlist users could claim spots that are reserved for users with pending waitlist offers, breaking the fairness guarantee of the batch system.

#### Step 1: Database Model Changes

**1. Add Waitlist Configuration Fields to `Event` Model:**

In `events/models/event.py`, add the following fields to the `Event` model:

```python
# Waitlist configuration
waitlist_batch_size = models.PositiveIntegerField(
    default=0,
    help_text="Number of users to notify per batch. 0 = notify everyone at once"
)
waitlist_time_window = models.DurationField(
    default=timedelta(hours=24),
    help_text="Time window for a batch to claim spots before moving to next batch"
)
waitlist_cutoff_date = models.DateTimeField(
    null=True,
    blank=True,
    db_index=True,
    help_text="After this date, all waitlist members are notified simultaneously"
)
waitlist_lottery_mode = models.BooleanField(
    default=False,
    help_text="If True, randomly select batch members instead of FIFO"
)
```

**2. Create `WaitlistOffer` Model:**

Create a new model in `events/models/event.py`:

```python
class WaitlistOffer(TimeStampedModel):
    """Represents a temporary offer for a waitlist user to claim a spot.

    Pending offers effectively 'reserve' spots at the event, preventing
    non-waitlist users from claiming them during the offer window.
    """

    class Status(models.TextChoices):
        PENDING = "pending", "Pending"
        CLAIMED = "claimed", "Claimed"
        EXPIRED = "expired", "Expired"

    event = models.ForeignKey(Event, on_delete=models.CASCADE, related_name="waitlist_offers")
    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, related_name="waitlist_offers")
    status = models.CharField(max_length=10, choices=Status.choices, default=Status.PENDING, db_index=True)
    expires_at = models.DateTimeField(db_index=True)
    notification_sent_at = models.DateTimeField(null=True, blank=True)
    claimed_at = models.DateTimeField(null=True, blank=True)
    batch_id = models.UUIDField(
        null=True,
        blank=True,
        db_index=True,
        help_text="Groups offers from the same batch notification"
    )

    class Meta:
        constraints = [
            models.UniqueConstraint(
                fields=["event", "user"],
                condition=models.Q(status="pending"),
                name="unique_pending_waitlist_offer"
            )
        ]
        indexes = [
            models.Index(fields=["event", "status", "expires_at"]),
            models.Index(fields=["batch_id", "status"]),
        ]
        ordering = ["-created_at"]
```

**3. Add Validation to Event Model:**

```python
def clean(self):
    super().clean()
    # ... existing validations ...

    # Validate waitlist time window
    if self.waitlist_time_window:
        if self.waitlist_time_window < timedelta(hours=1):
            raise DjangoValidationError(
                {"waitlist_time_window": "Time window must be at least 1 hour"}
            )
        if self.waitlist_time_window > timedelta(days=7):
            raise DjangoValidationError(
                {"waitlist_time_window": "Time window cannot exceed 7 days"}
            )

    # Validate cutoff date is before event start
    if self.waitlist_cutoff_date and self.waitlist_cutoff_date > self.start:
        raise DjangoValidationError(
            {"waitlist_cutoff_date": "Cutoff date must be before event start"}
        )
```

#### Step 2: Core Waitlist Service

Create `events/service/waitlist_service.py`:

```python
from uuid import uuid4
from django.db import transaction
from django.utils import timezone
from events.models import Event, EventWaitList, WaitlistOffer, Ticket, EventRSVP
from typing import List
import random

def get_available_spots(event: Event) -> int:
    """Calculate how many spots are available for NEW waitlist offers.

    This counts: max_attendees - (current_attendees + pending_offers)
    Pending offers effectively "reserve" spots.
    """
    if event.max_attendees == 0:
        return 9999  # Effectively infinite

    attendee_count = event.attendee_count
    pending_offers = WaitlistOffer.objects.filter(
        event=event,
        status=WaitlistOffer.Status.PENDING
    ).count()

    # Pending offers reserve spots, so they count toward capacity
    return max(0, event.max_attendees - attendee_count - pending_offers)

def get_actual_capacity(event: Event) -> int:
    """Calculate the TRUE available capacity including pending offers.

    This is used by the AvailabilityGate to determine if users
    WITH pending offers can register.
    """
    if event.max_attendees == 0:
        return 9999

    attendee_count = event.attendee_count
    # Don't count pending offers - they represent available spots for offer holders
    return max(0, event.max_attendees - attendee_count)

@transaction.atomic
def process_waitlist_for_event(event_id: UUID) -> dict:
    """Main entry point for processing waitlist when spots become available."""
    event = Event.objects.select_for_update().get(id=event_id)

    # Check if we're past cutoff date
    now = timezone.now()
    past_cutoff = event.waitlist_cutoff_date and now >= event.waitlist_cutoff_date

    available_spots = get_available_spots(event)
    if available_spots <= 0:
        return {"status": "no_spots_available", "offers_created": 0}

    # Determine batch size
    if past_cutoff or event.waitlist_batch_size == 0:
        batch_size = available_spots  # Notify everyone
    else:
        batch_size = min(event.waitlist_batch_size, available_spots)

    # Get waitlist entries that don't have pending offers
    waitlist_qs = EventWaitList.objects.filter(event=event) \
        .exclude(
            user__waitlist_offers__event=event,
            user__waitlist_offers__status=WaitlistOffer.Status.PENDING
        ) \
        .select_related('user')

    if event.waitlist_lottery_mode and not past_cutoff:
        # Random selection (lottery mode)
        all_entries = list(waitlist_qs)
        selected_entries = random.sample(all_entries, min(batch_size, len(all_entries)))
    else:
        # FIFO selection
        selected_entries = list(waitlist_qs.order_by('created_at')[:batch_size])

    if not selected_entries:
        return {"status": "no_eligible_users", "offers_created": 0}

    # Create offers for selected users
    batch_id = uuid4()
    offers_created = 0
    for waitlist_entry in selected_entries:
        create_offer(event, waitlist_entry.user, batch_id)
        offers_created += 1

    return {
        "status": "success",
        "offers_created": offers_created,
        "batch_id": str(batch_id),
        "past_cutoff": past_cutoff
    }

def create_offer(event: Event, user: RevelUser, batch_id: UUID) -> WaitlistOffer:
    """Create a waitlist offer for a user."""
    offer = WaitlistOffer.objects.create(
        event=event,
        user=user,
        expires_at=timezone.now() + event.waitlist_time_window,
        batch_id=batch_id
    )

    # Trigger notification task
    from events.tasks import send_waitlist_offer_notification_task
    send_waitlist_offer_notification_task.delay(str(offer.id))

    return offer

def claim_offer(offer: WaitlistOffer) -> bool:
    """Mark an offer as claimed and remove user from waitlist."""
    if offer.status != WaitlistOffer.Status.PENDING:
        return False

    if timezone.now() > offer.expires_at:
        offer.status = WaitlistOffer.Status.EXPIRED
        offer.save()
        return False

    offer.status = WaitlistOffer.Status.CLAIMED
    offer.claimed_at = timezone.now()
    offer.save()

    # Remove from waitlist
    EventWaitList.objects.filter(event=offer.event, user=offer.user).delete()

    return True
```

#### Step 3: Eligibility Service Integration

**Create `ActiveWaitlistOfferGate` in `events/service/event_manager.py`:**

```python
class ActiveWaitlistOfferGate(BaseEligibilityGate):
    """Gate #0: Grants access if the user has an active waitlist offer.

    This gate MUST come first to allow users with pending offers to
    bypass the AvailabilityGate's capacity check.
    """

    def check(self) -> EventUserEligibility | None:
        """Check if user has an active waitlist offer."""
        active_offer = getattr(self.handler, 'active_waitlist_offer', None)
        if active_offer and active_offer.event_id == self.event.id:
            # Check if offer has expired
            if active_offer.expires_at > timezone.now():
                return EventUserEligibility(allowed=True, event_id=self.event.id)
        return None
```

**Update `AvailabilityGate` to use the correct capacity calculation:**

```python
class AvailabilityGate(BaseEligibilityGate):
    """Gate #7: Checks if the event has space available for another attendee."""

    def check(self) -> EventUserEligibility | None:
        """Check if the event has space available for another attendee."""
        if self.event.max_attendees == 0 or self.handler.overrides_max_attendees():
            return None

        # Use actual capacity which includes space for pending offers
        from events.service.waitlist_service import get_actual_capacity
        available_capacity = get_actual_capacity(self.event)

        if available_capacity <= 0:
            return EventUserEligibility(
                allowed=False,
                event_id=self.event.id,
                reason=_(Reasons.EVENT_IS_FULL),
                next_step=NextStep.JOIN_WAITLIST if self.event.waitlist_open else None,
            )
        return None
```

**Update `EligibilityService.__init__`:**

```python
def __init__(self, user: RevelUser, event: models.Event) -> None:
    # ... existing code ...

    # Prefetch active waitlist offer
    self.active_waitlist_offer = user.waitlist_offers.filter(
        event=event,
        status=WaitlistOffer.Status.PENDING,
        expires_at__gt=timezone.now()
    ).first()

    # ... rest of existing code ...
```

**Update `EligibilityService.ELIGIBILITY_GATES`:**

```python
ELIGIBILITY_GATES = [
    ActiveWaitlistOfferGate,  # NEW: Check first!
    PrivilegedAccessGate,
    EventStatusGate,
    RSVPDeadlineGate,
    InvitationGate,
    MembershipGate,
    QuestionnaireGate,
    AvailabilityGate,
    TicketSalesGate,
]
```

#### Step 4: Triggers (Signals)

Create or update `events/signals.py`:

```python
from django.db.models.signals import post_save, post_delete
from django.dispatch import receiver
from events.models import Ticket, EventRSVP, Event
from events.tasks import process_waitlist_for_event_task

@receiver(post_save, sender=Ticket)
def handle_ticket_change(sender, instance: Ticket, created: bool, **kwargs):
    """Trigger waitlist processing when tickets are cancelled.

    Note: attendee_count is NOT decremented here. It will only decrease
    when offers expire without being claimed.
    """
    if not created and instance.status == Ticket.Status.CANCELLED:
        process_waitlist_for_event_task.delay(str(instance.event_id))

@receiver(post_save, sender=EventRSVP)
def handle_rsvp_change(sender, instance: EventRSVP, created: bool, **kwargs):
    """Trigger waitlist processing when RSVPs change from YES.

    Note: attendee_count is NOT decremented here. It will only decrease
    when offers expire without being claimed.
    """
    if not created:
        # We need to detect if status changed from YES to something else
        # This is tricky without update_fields; consider using django-simple-history
        # For now, trigger on any status change to/from YES
        if instance.status != EventRSVP.Status.YES:
            process_waitlist_for_event_task.delay(str(instance.event_id))

@receiver(post_save, sender=Event)
def handle_capacity_increase(sender, instance: Event, created: bool, **kwargs):
    """Trigger waitlist processing when max_attendees increases."""
    if not created and instance.waitlist_open:
        # This triggers on any Event save when waitlist is open
        # Ideally we'd only trigger if max_attendees increased
        process_waitlist_for_event_task.delay(str(instance.id))
```

#### Step 5: Celery Tasks

Add to `events/tasks.py`:

```python
@shared_task
def process_waitlist_for_event_task(event_id: str) -> dict:
    """Celery task to process waitlist for an event."""
    from events.service.waitlist_service import process_waitlist_for_event
    return process_waitlist_for_event(UUID(event_id))

@shared_task
def send_waitlist_offer_notification_task(offer_id: str) -> dict:
    """Send notification to user about their waitlist offer."""
    offer = WaitlistOffer.objects.select_related('user', 'event').get(id=offer_id)

    subject = f"Spot Available: {offer.event.name}"
    context = {
        "user": offer.user,
        "event": offer.event,
        "offer": offer,
        "expires_at": offer.expires_at,
    }

    body = render_to_string("events/emails/waitlist_offer.txt", context)
    html_body = render_to_string("events/emails/waitlist_offer.html", context)

    site_settings = SiteSettings.get_solo()
    recipient = to_safe_email_address(offer.user.email, site_settings=site_settings)

    email_msg = EmailMultiAlternatives(
        subject=subject,
        body=body,
        from_email=settings.DEFAULT_FROM_EMAIL,
        bcc=[recipient],
    )

    if html_body:
        email_msg.attach_alternative(html_body, "text/html")

    # Attach event ICS
    ics_bytes = offer.event.ics()
    email_msg.attach("event.ics", ics_bytes, "text/calendar")

    email_msg.send(fail_silently=False)

    # Log email
    email_log = EmailLog(to=recipient, subject=subject)
    email_log.set_body(body=body)
    if html_body:
        email_log.set_html(html_body=html_body)
    email_log.save()

    # Update offer
    offer.notification_sent_at = timezone.now()
    offer.save(update_fields=['notification_sent_at'])

    return {"status": "sent", "offer_id": str(offer.id)}

@shared_task(name="events.expire_waitlist_offers")
def expire_waitlist_offers() -> dict:
    """Periodic task to expire old waitlist offers and process next batches.

    When offers expire, the spots they were holding become available for
    the next batch or for general registration (if no more waitlist).
    """
    now = timezone.now()

    # Find expired offers
    expired_offers = WaitlistOffer.objects.filter(
        status=WaitlistOffer.Status.PENDING,
        expires_at__lt=now
    ).select_related('event')

    # Update status
    expired_count = expired_offers.update(status=WaitlistOffer.Status.EXPIRED)

    if expired_count > 0:
        # Get unique event IDs
        event_ids = set(expired_offers.values_list('event_id', flat=True))

        # Trigger next batch for each affected event
        # This will check if there are more people in waitlist
        # If not, the spots become available for general registration
        for event_id in event_ids:
            process_waitlist_for_event_task.delay(str(event_id))

    return {
        "expired_count": expired_count,
        "events_processed": len(event_ids) if expired_count > 0 else 0
    }
```

**Configure Celery Beat:**

In Celery configuration, schedule the expiration task to run every 1-5 minutes:

```python
beat_schedule = {
    'expire-waitlist-offers': {
        'task': 'events.expire_waitlist_offers',
        'schedule': crontab(minute='*/5'),  # Every 5 minutes
    },
}
```

#### Step 6: Update EventManager._assert_capacity

Modify `events/service/event_manager.py` (lines 564-595) to account for pending waitlist offers:

**Current behavior:** The method only counts actual attendees (tickets/RSVPs).

**Required change:** Include pending waitlist offers in the capacity calculation to prevent non-waitlist users from claiming reserved spots.

```python
def _assert_capacity(self, use_tickets: bool, tier: TicketTier | None) -> None:
    """Raises if the event has no more available attendee slots.

    This method now accounts for pending waitlist offers, which effectively
    reserve spots during their time window.
    """
    if self.event.max_attendees == 0 or self.eligibility_service.overrides_max_attendees():
        return

    if use_tickets:
        count = Ticket.objects.select_for_update().filter(event=self.event).values("user_id").distinct().count()
        if not tier:
            raise ValueError("Tier must be provided for ticket counts.")
        if tier.total_quantity and tier.quantity_sold >= tier.total_quantity:
            raise UserIsIneligibleError(
                message="Tier is sold out.",
                eligibility=EventUserEligibility(
                    allowed=False,
                    event_id=self.event.id,
                    next_step=NextStep.JOIN_WAITLIST if self.event.waitlist_open else None,
                    reason=_(Reasons.SOLD_OUT),
                ),
            )
    else:
        count = EventRSVP.objects.select_for_update().filter(event=self.event, status=EventRSVP.Status.YES).count()

    # NEW: Add pending waitlist offers to the count
    # Pending offers reserve spots for their holders
    from events.models import WaitlistOffer
    pending_offers_count = WaitlistOffer.objects.select_for_update().filter(
        event=self.event,
        status=WaitlistOffer.Status.PENDING
    ).count()

    # Check total capacity including reserved spots
    if count + pending_offers_count >= self.event.max_attendees:
        raise UserIsIneligibleError(
            message="Event is full.",
            eligibility=EventUserEligibility(
                allowed=False,
                event_id=self.event.id,
                next_step=NextStep.JOIN_WAITLIST if self.event.waitlist_open else None,
                reason=_(Reasons.EVENT_IS_FULL),
            ),
        )
```

**Why this matters:** Without this update, non-waitlist users could claim spots that are reserved for users with pending waitlist offers, breaking the fairness guarantee of the batch system.

**Update EventManager methods to claim offers:**

```python
@transaction.atomic
def create_ticket(self, tier: TicketTier, bypass_eligibility_checks: bool = False, price_override: Decimal | None = None) -> Ticket | str:
    # ... existing code ...

    # After successful ticket creation, claim any active offer
    active_offer = self.user.waitlist_offers.filter(
        event=self.event,
        status=WaitlistOffer.Status.PENDING
    ).first()

    if active_offer:
        from events.service.waitlist_service import claim_offer
        claim_offer(active_offer)

    return ticket_or_url

@transaction.atomic
def rsvp(self, answer: EventRSVP.Status, bypass_eligibility_checks: bool = False) -> EventRSVP:
    # ... existing code ...

    # After successful RSVP, claim any active offer
    if answer == EventRSVP.Status.YES:
        active_offer = self.user.waitlist_offers.filter(
            event=self.event,
            status=WaitlistOffer.Status.PENDING
        ).first()

        if active_offer:
            from events.service.waitlist_service import claim_offer
            claim_offer(active_offer)

    return rsvp
```

---

### Testing Strategy

1. **Unit Tests:**
   - Test `get_available_spots()` with various scenarios
   - Test `get_actual_capacity()` distinction from available spots
   - Test FIFO vs lottery selection logic
   - Test cutoff date behavior
   - Test offer expiration and claiming
   - Test `ActiveWaitlistOfferGate` in isolation
   - Test `_assert_capacity()` with pending offers

2. **Integration Tests:**
   - **Spot Reservation:** Test that cancellation + offers creation doesn't immediately free up spots
   - **Capacity Tracking:** Test that pending offers reserve spots correctly
   - Test full flow: ticket cancellation → batch notification → claim
   - Test batch expiration → next batch notification
   - Test cutoff date triggers full notification
   - Test race conditions with multiple spots and concurrent claims
   - Test lottery mode randomness
   - **Critical:** Test that non-waitlist users cannot register while offers are pending
   - **Critical:** Test that `_assert_capacity()` correctly blocks registration when spots are reserved

3. **Signal Tests:**
   - Test signals trigger waitlist processing correctly
   - Test idempotency of signals

4. **Celery Task Tests:**
   - Test scheduled expiration task
   - Test notification task
   - Test that expired offers trigger next batch correctly
   - Mock email sending

---

### Migration Plan

1. Create migration for new fields and model
2. Set sensible defaults for existing events with `waitlist_open=True`
3. No data migration needed (waitlist offers start fresh)
4. Update admin to expose new fields
5. Update API schemas to include new fields

---

### Admin Integration

Update `events/admin.py`:

```python
@admin.register(models.WaitlistOffer)
class WaitlistOfferAdmin(ModelAdmin):
    list_display = ['id', 'event_link', 'user_link', 'status', 'expires_at', 'batch_id', 'created_at']
    list_filter = ['status', 'event']
    search_fields = ['user__email', 'event__name', 'batch_id']
    readonly_fields = ['created_at', 'notification_sent_at', 'claimed_at', 'batch_id']
```

Update Event admin fieldsets to include new waitlist fields.

---

### API Changes

Update `events/schema.py` to include new fields in event schemas:

```python
class EventCreateIn(Schema):
    # ... existing fields ...
    waitlist_batch_size: int = 0
    waitlist_time_window: timedelta = timedelta(hours=24)
    waitlist_cutoff_date: datetime | None = None
    waitlist_lottery_mode: bool = False
```

biagiodistefano:
Closes #2.

## Summary

Replaces the passive `EventWaitList` with a configurable, batched, time-limited offer flow. When capacity opens, the system selects a batch of waitlist users (FIFO or lottery), notifies them with an expiry deadline, and **reserves their spots** during the time window so non-waitlist users can't squeeze in.

Opt-in per event: setting `Event.waitlist_time_window` to a non-null duration turns the feature on; otherwise the event keeps today's passive behavior.

Design doc: `docs/superpowers/specs/2026-05-19-advanced-waitlist-design.md`. Plan: `docs/superpowers/plans/2026-05-19-advanced-waitlist.md`. (Both local-only — `docs/superpowers/` is gitignored.)

## What it does

- **Data model** — five new `Event` fields (`waitlist_time_window`, `waitlist_batch_size`, `waitlist_cutoff_date`, `waitlist_cutoff_window`, `waitlist_lottery_mode`) + new `WaitlistOffer` model with partial unique constraint on `(event, user)` where `status='pending'`.
- **Reservation semantics** — both `AvailabilityGate` (prefetched) and `EventManager._assert_capacity` (row-locked) include pending unexpired offers in their counting; a user's own active offer is *not* double-counted.
- **Selection** — `process_waitlist_for_event()` in `events/service/waitlist_service.py` is the single entrypoint. FIFO by default, lottery via `random.sample`, one-shot all-hands batch on cutoff, idempotent via Event row lock.
- **Triggers** — explicit `enqueue_waitlist_processing(event_id)` calls wired into every capacity-freeing path: user/admin ticket cancel, Stripe refund webhook, cancel_checkout, RSVP YES→non-YES (user + admin), `update_event` capacity increase, `update_event_status` un-cancel, leave_waitlist, manual offer revoke, hourly expire sweeper.
- **Claim** — implicit on successful registration (RSVP YES or non-cancelled ticket creation, including PENDING Stripe tickets) in `EventManager.rsvp` and `BatchTicketService._create_tickets`.
- **Notification** — uses the existing `notification_requested` signal with `WAITLIST_SPOT_AVAILABLE`. Context includes event TZ-formatted absolute time + humanized relative time + cutoff flag. Templates updated for all four channels (email, in-app, telegram).
- **Admin surface** — new `EventAdminWaitlistOffersController` with settings GET/PATCH, offers list, revoke, **reactivate**, and **manual create** endpoints. Public `EventSchema` gains `seats_held`. `EventUserEligibility` gains `pending_offers_count`, `next_batch_at`, `waitlist_position`, `active_offer_expires_at`. New `Reasons.SPOTS_RESERVED_FOR_WAITLIST` / `ON_WAITLIST_WAITING_FOR_BATCH` and `NextStep.CLAIM_WAITLIST_OFFER`.
- **Django admin** — `WaitlistOfferAdmin` registered, inline on `EventAdmin`, "Waitlist" fieldset, Unfold sidebar entry.
- **Hardened join_waitlist** — runs full eligibility check (blacklist, invitation, membership, etc.); **409 Conflict** when event has free capacity (FE refreshes and registers directly); enqueues processing on join so stale-page joiners get a same-second offer if a seat freed between page-load and click.

## Notable decisions (full rationale in the spec)

- **`attendee_count` is unchanged** — denormalized field still tracks committed attendees only; reservation lives entirely in the offers table.
- **One-shot cutoff** — after the cutoff_date, one final notify-all batch fires, then `AvailabilityGate` resumes normal behavior. Idempotency marker is `is_cutoff_batch=True` on the offer row.
- **Explicit triggers, one exception** — `EventWaitList.post_delete` is the only signal-based hook (rows are deleted from too many paths to wire inline). The signal is documented inline at the receiver and at the admin delete-entry site.
- **Hourly Beat sweeper** — `expire_waitlist_offers_task` flips PENDING→EXPIRED and enqueues next-batch processing. Read paths defensively filter `expires_at > now` so the slower cadence doesn't affect capacity correctness.
- **PostgreSQL nit** — `FOR UPDATE` + `DISTINCT` is rejected, so we materialize and de-dup in Python.
- **Eligibility-aware join** — non-eligible users (blacklisted, no invitation, etc.) can't sit on the waitlist forever waiting for an offer they can never claim.

## Test plan

- [ ] `make check` green (ruff, mypy strict, migrations, i18n, file-length).
- [ ] `make test` green — full suite was **4601 passed / 0 failed** in the last linear run (after the test_tasks.py and test_signals.py file/package rename fixes).
- [ ] Targeted manual verification on a staging event:
  - [ ] Enable advanced waitlist on a test event, fill to capacity.
  - [ ] Cancel a ticket → confirm a batch of `batch_size` offers fires + notifications arrive.
  - [ ] Hit `/event-admin/{id}/waitlist-offers` to inspect.
  - [ ] Have an offer holder register → confirm `CLAIMED`, EventWaitList row drops, next batch fires if more spots open.
  - [ ] Try to register as a non-offer holder while batch pending → `SPOTS_RESERVED_FOR_WAITLIST` reason.
  - [ ] Wait past expiry → confirm hourly sweeper flips status + fires next batch (or trigger task manually).
  - [ ] Set a cutoff date in the past → confirm one-shot all-hands batch.
  - [ ] Toggle `waitlist_open` off via admin endpoint → confirm pending offers REVOKED.
  - [ ] Admin reactivate / manual-create / revoke endpoints behave as expected.
  - [ ] Join waitlist when event has capacity → confirm 409.
  - [ ] Join waitlist as a blacklisted user → confirm ineligibility.
- [ ] Frontend regen: `pnpm generate:api` in `revel-frontend/` to pick up the new schemas.

## Migration summary

Three new migrations on `events/`:
- `0073_event_waitlist_batch_size_..._and_more` — five new fields on Event.
- `0074_waitlistoffer` — new model.
- `0075_waitlist_periodic_task` — registers the hourly Beat schedule.

All backwards-compatible. Existing events stay in legacy passive mode (`waitlist_time_window=NULL`).

## Summary by CodeRabbit

* **New Features**
  * Advanced waitlist offers: batched/lottery selection, cutoff handling, timed expirations, revocation, reactivation, and per-offer notifications
  * Admin UI & APIs to view/update waitlist settings and manage offers; manual offer creation/revocation
  * Automatic claiming of offers when RSVPing or purchasing; capacity accounting includes reserved offers

* **Chores**
  * DB migrations and periodic tasks for offer expiry and nudging

* **Tests**
  * Extensive unit, controller, integration and task coverage for waitlist behaviors and edge cases

* **Docs**
  * New and updated architecture and notification docs for advanced waitlists

[Review Change Stack](https://app.coderabbit.ai/change-stack/letsrevel/revel-backend/pull/426?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


biagiodistefano:
### Code review

Reviewed alongside Copilot and CodeRabbit. Both bots posted thorough inline comments — this comment focuses on findings they missed and the highest-impact items worth elevating.

#### Findings missed by Copilot and CodeRabbit

**1. `BatchTicketService._assert_event_capacity` does not count pending waitlist offers — breaks the reservation guarantee for ticket events**

The PR offer-aware-ifies `EventManager._assert_capacity` (RSVP path) but `BatchTicketService._assert_event_capacity` is the actual hot path for ticket purchases (`BatchTicketService.create_batch` → `_assert_event_capacity` → `_create_tickets`). It still computes `available = effective_cap - non_cancelled_ticket_count` with no `WaitlistOffer` accounting. A non-offer-holder can buy a ticket while a batch of offers is outstanding, exactly the scenario this PR is designed to prevent.

https://github.com/letsrevel/revel-backend/blob/09a5f2f86f4394d1360d61fe6b5a68237645ae56/src/events/service/batch_ticket_service.py#L335-L368

**2. Admin `delete_rsvp` doesn't enqueue waitlist processing**

`create_rsvp` and `update_rsvp` both enqueue on YES → non-YES transitions (commit `9d88bf3`), but `delete_rsvp` just calls `rsvp.delete()` and returns. Deleting a YES RSVP frees a seat without triggering the next waitlist batch.

https://github.com/letsrevel/revel-backend/blob/09a5f2f86f4394d1360d61fe6b5a68237645ae56/src/events/controllers/event_admin/rsvps.py#L127-L137

**3. Manual offer creation dispatches notification without `transaction.on_commit`**

Every other dispatch site in the PR correctly wraps the `.delay()` in `transaction.on_commit(lambda: ...)`. `create_waitlist_offer` (and `reactivate_waitlist_offer`) call `send_waitlist_offer_notification_task.delay(str(offer.id))` directly after the create/save. If a future caller wraps these endpoints in an outer atomic block and rolls back, the notification fires for an offer that never persisted.

https://github.com/letsrevel/revel-backend/blob/09a5f2f86f4394d1360d61fe6b5a68237645ae56/src/events/controllers/event_admin/waitlist_offers.py#L147-L194

**4. `AvailabilityGate`'s "zero additional DB queries" contract is now broken**

Class docstring says: *"This is a preliminary capacity check using prefetched data (zero additional DB queries)."* The new `_earliest_pending_expiry` and `_get_waitlist_position` helpers both issue fresh queries every time the gate is the blocker. Either prefetch/annotate these onto the event in `EligibilityService` (matches the existing pattern for `pending_waitlist_offer_count`) or relax the docstring.

https://github.com/letsrevel/revel-backend/blob/09a5f2f86f4394d1360d61fe6b5a68237645ae56/src/events/service/event_manager/gates.py#L653-L680

**5. `WaitlistOfferSchema` includes `created_at` (CLAUDE.md: "Don't include `created_at`/`updated_at` in response schemas unless specifically needed")**

[revel-backend/CLAUDE.md](https://github.com/letsrevel/revel-backend/blob/09a5f2f86f4394d1360d61fe6b5a68237645ae56/CLAUDE.md) is explicit about omitting timestamps from ModelSchema responses. Ordering is already handled server-side via `Meta.ordering = ["-created_at"]`.

https://github.com/letsrevel/revel-backend/blob/09a5f2f86f4394d1360d61fe6b5a68237645ae56/src/events/schema/waitlist.py#L34-L53

#### Highest-impact items from Copilot/CodeRabbit worth elevating

These are already inline-commented, but they're the ones I'd block on before merge:

- **Cutoff branch over-reserves seats** (Copilot, `waitlist_service.py:87`) — `selected = [w.user for w in waitlist_qs]` creates offers for *every* remaining waitlist user with no `available` cap. Since pending offers count toward capacity, `len(selected) > available` over-reserves and could leave even offer-holders unable to register.
- **`reactivate_waitlist_offer` accepts CLAIMED state** (Copilot/CodeRabbit, `waitlist_offers.py:122-124`) — current guard only rejects PENDING. A CLAIMED offer can be flipped back to PENDING, breaking the lifecycle invariant. Whitelist `{EXPIRED, REVOKED}` instead.
- **`_assert_capacity` doesn't `select_for_update` on the Event row** (CodeRabbit) — `process_waitlist_for_event` locks the event before reading offer counts; `_assert_capacity` does not, so a concurrent RSVP/ticket flow can read `pending_offers=0` racing against batch creation and let both commit.
- **TOCTOU race on uniqueness writes** (Copilot/CodeRabbit, multiple sites) — `exists()`-then-`create()` on `WaitlistOffer` (manual create + reactivate) and `EventWaitList` (`join_waitlist`) will surface 500 IntegrityError instead of intended 409/200 under concurrency. CLAUDE.md prescribes `common.utils.get_or_create_with_race_protection` for exactly this.
- **`revoke_all_pending_offers` docstring contradicts filter** (Copilot + Claude) — docstring promises "already-expired rows untouched"; filter is `status=PENDING` only. Either add `expires_at__gt=now` or rewrite the docstring.

#### Tests / docs

CodeRabbit's nitpicks on test tightening (`assert_called_once_with` over `assert_called_with` in `test_enqueue_wires.py`, asserting specific blocker reason in the lifecycle integration test, edge-case tests for `resolve_seats_held`) are all worth picking up — none blocking.

src/events/controllers/event_admin/tickets.py
```diff
@@ -313,6 +314,8 @@ def mark_ticket_refunded(
                 ticket.payment.refunded_at = timezone.now()
                 ticket.payment.save(update_fields=["status", "refund_amount", "refund_status", "refunded_at"])
 
+            enqueue_waitlist_processing(ticket.event_id)
```
biagiodistefano, line 317: Maybe this should be on a on_commit()?

src/events/controllers/event_admin/tickets.py
```diff
@@ -366,6 +369,8 @@ def cancel_ticket(
                 ]
             )
 
+            enqueue_waitlist_processing(ticket.event_id)
```
biagiodistefano, line 372: Maybe this should be on a on_commit()?

src/events/models/event.py
```diff
@@ -551,6 +610,47 @@ def __str__(self) -> str:
...
 
+class WaitlistOffer(TimeStampedModel):
+    """Temporary offer granting a user the right to claim a reserved spot.
+
+    Pending unexpired offers count toward the event's capacity in both the
+    AvailabilityGate (prefetched) and EventManager._assert_capacity (row-locked).
+    See docs/superpowers/specs/2026-05-19-advanced-waitlist-design.md.
+    """
+
+    class Status(models.TextChoices):
```
biagiodistefano: This name is too generic and high risk of collision in schema generation. Change to WaitlistOfferStatus

src/events/controllers/event_admin/waitlist_offers.py
```diff
@@ -0,0 +1,212 @@
...
+        offer.notified_at = None
+        try:
+            with transaction.atomic():
+                offer.save(update_fields=["status", "expires_at", "claimed_at", "notified_at"])
+        except IntegrityError as exc:
+            # Lost the race: another writer landed a PENDING offer for this
+            # (event, user) between the conflict check above and the save.
+            raise HttpError(409, "User already has a pending offer for this event.") from exc
+        offer_id_str = str(offer.id)
+        transaction.on_commit(lambda: send_waitlist_offer_notification_task.delay(offer_id_str))
```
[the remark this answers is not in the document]
biagiodistefano: Verify if this is valid and if yes, if fixed

src/events/models/event.py
```diff
@@ -246,6 +246,39 @@ class EventStatus(models.TextChoices):
...
+        help_text=(
+            "After this datetime, batching stops. One final all-hands batch fires, "
+            "then AvailabilityGate behaves normally."
+        ),
+    )
+    waitlist_cutoff_window = models.DurationField(
+        null=True,
+        blank=True,
+        help_text=("Time window for the cutoff all-hands batch. Falls back to waitlist_time_window when null."),
+    )
```
biagiodistefano: This doesn't make sense. After the cut-off date, everyone gets notified, there is no point in having a limited time window for users to respond. Nuke this option and all the logic/schema that uses it.

src/events/schema/rsvp.py
```diff
@@ -72,11 +73,30 @@ class WaitlistEntrySchema(ModelSchema):
...
+        material.
+        """
+        return (
+            WaitlistOffer.objects.filter(
+                event_id=obj.event_id,
+                user_id=obj.user_id,
+                status=WaitlistOffer.Status.PENDING,
+            )
+            .order_by("-created_at")
+            .first()
```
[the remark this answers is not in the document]
biagiodistefano: Double check if correct. If confirmed, fix.

src/events/service/event_manager/types.py
```diff
@@ -20,6 +20,10 @@ class EventUserEligibility(BaseModel):
     questionnaires_failed: list[uuid.UUID] | None = None
     retry_on: datetime.datetime | None = None
     missing_profile_fields: list[str] | None = None
+    pending_offers_count: int | None = None
+    next_batch_at: datetime.datetime | None = None
```
biagiodistefano: should be a pydantic's aware datetime. Resolve to user's timezone if available otherwise to system's timezone.

src/events/service/event_manager/types.py
```diff
@@ -20,6 +20,10 @@ class EventUserEligibility(BaseModel):
...
+    waitlist_position: int | None = None
+    active_offer_expires_at: datetime.datetime | None = None
```
biagiodistefano: should be a pydantic's aware datetime. Resolve to user's timezone if available otherwise to system's timezone.

src/events/service/event_manager/types.py
```diff
@@ -20,6 +20,10 @@ class EventUserEligibility(BaseModel):
     questionnaires_failed: list[uuid.UUID] | None = None
     retry_on: datetime.datetime | None = None
```
biagiodistefano: Unrelated but since we're fixing the other ones: should be a pydantic's aware datetime. Resolve to user's timezone if available otherwise to system's timezone.

src/events/service/cancellation_service.py
```diff
@@ -406,6 +407,8 @@ def cancel_ticket_by_user(
             quantity_sold=F("quantity_sold") - 1
         )
 
+        enqueue_waitlist_processing(locked_ticket.event_id)
```
biagiodistefano, line 410: Again: should it be on_commit()?

src/events/controllers/event_admin/tickets.py
biagiodistefano: `enqueue_waitlist_processing` already wraps `transaction.on_commit` internally — see [waitlist_service.py:156-162](https://github.com/letsrevel/revel-backend/blob/b870a0d/src/events/service/waitlist_service.py#L156-L163). Calling it from within an atomic block is on_commit-safe by design: the lambda fires after the outermost commit, or never if the transaction rolls back. No fix needed here. (Will add a clarifying line to the helper docstring so this is less surprising.)

src/events/controllers/event_admin/tickets.py
biagiodistefano: `enqueue_waitlist_processing` already wraps `transaction.on_commit` internally — see [waitlist_service.py:156-162](https://github.com/letsrevel/revel-backend/blob/b870a0d/src/events/service/waitlist_service.py#L156-L163). Calling it from within an atomic block is on_commit-safe by design: the lambda fires after the outermost commit, or never if the transaction rolls back. No fix needed here. (Will add a clarifying line to the helper docstring so this is less surprising.)

src/events/service/cancellation_service.py
biagiodistefano: `enqueue_waitlist_processing` already wraps `transaction.on_commit` internally — see [waitlist_service.py:156-162](https://github.com/letsrevel/revel-backend/blob/b870a0d/src/events/service/waitlist_service.py#L156-L163). Calling it from within an atomic block is on_commit-safe by design: the lambda fires after the outermost commit, or never if the transaction rolls back. No fix needed here. (Will add a clarifying line to the helper docstring so this is less surprising.)

biagiodistefano:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance across the delta since the previous review ([09a5f2f...5029f35](https://github.com/letsrevel/revel-backend/compare/09a5f2f...5029f35)).

Issue #470, opened 2026-06-01
feat(events): persist optional cancellation reason on event + surface in cancellation notification and detail schema


biagiodistefano:
## Context

Frontend issue: [letsrevel/revel-frontend#391](https://github.com/letsrevel/revel-frontend/issues/391)

When an organizer cancels an event, they should be able to attach an **optional reason** that is:
1. Persisted on the event
2. Included in the `EVENT_CANCELLED` notification payload (in-app / email / telegram)
3. Returned in the event response schemas so the frontend can display it on the event page

Most of the notification plumbing is already scaffolded — only the persistence + endpoint + schema bits are missing.

## Current state

**Already in place:**
- `notifications/context_schemas.py:184` — `EventCancelledContext.cancellation_reason: t.NotRequired[str]` declared
- `notifications/signals/event.py:68-69` — signal handler already conditionally appends the field via `hasattr(instance, "cancellation_reason")` (defensive scaffold awaiting the field)
- Templates already render the reason conditionally:
  - `notifications/templates/notifications/in_app/event_cancelled.md`
  - `notifications/templates/notifications/email/event_cancelled.{html,txt}`
  - `notifications/templates/notifications/telegram/event_cancelled.md`

**Missing:**
- `Event.cancellation_reason` field
- Migration
- Field on `EventBaseSchema` / `EventDetailSchema`
- Endpoint mechanism to send the reason alongside the status transition
- Service support in `events/service/event_update_service.update_status`

### 1. Model (`events/models/event.py`)
Add a `cancellation_reason` field on `Event`, mirroring `Ticket.cancellation_reason` (`events/models/ticket.py:693`):

```python
cancellation_reason = models.CharField(max_length=500, blank=True, default=\"\")
```

Open question: 500 chars (consistent with ticket) vs `TextField` (unbounded). Organizers may want more room than a ticket reason; happy to bump to 1000 or use `TextField` if preferred. **Default: 500 for consistency.**

### 2. Migration
Single AddField migration — no data backfill required (blank default).

### 3. Schemas (`events/schema/event.py`)
Add `cancellation_reason: str | None = None` to `EventBaseSchema` (so it's available on both list and detail). The empty-string default at the DB level can be normalized to `None` in the schema if we prefer (resolver), or just surfaced as-is — minor.

### 4. Endpoint (`events/controllers/event_admin/core.py:108`)
Today's signature:
```python
@route.post(\"/actions/update-status/{status}\", ...)
[...]

PR #472, opened 2026-06-02
fix(events): repair event-cancelled notification + gate cancellation reason to attendees

biagiodistefano:
## Summary

Two related fixes around event cancellation, surfaced by an e2e smoke test of the #470 feature.

### 1. Event-cancellation notifications were silently never delivered 🐛

`EventCancelledContext` declares `refund_available: bool` as **required**, but `_handle_event_cancelled` only ever set `refund_info` — never `refund_available`. So `create_notification` → `validate_notification_context` raised `ValueError: Missing required context keys ... {'refund_available'}`, which `handle_notification_request` catches and swallows by design. Net effect: **no event-cancellation notification (in-app / email / telegram) was ever created.**

This is a latent pre-existing bug (not introduced by #470) — the #470 signal tests mocked `notification_requested.send`, so they never exercised the validator.

**Fix:** populate `refund_available` (= `event.requires_ticket`, consistent with how `refund_info` is gated) + a regression test that drives the real `notification_requested → create_notification → validate_notification_context` path and asserts a `Notification` row is actually persisted.

### 2. `cancellation_reason` is now disclosed only to attendees 🔒

The reason is addressed to people who were going to attend, so it shouldn't be readable by every user who can see the event. Added `Event.can_user_see_cancellation_reason()` mirroring the **ATTENDEES tier** of `can_user_see_address` (ticket holders + confirmed `YES` RSVPs, plus staff/owners and superusers; a merely-invited or anonymous user does **not** qualify), and gated `EventBaseSchema.resolve_cancellation_reason` on it.

Empty reasons short-circuit **before** any access query, so the common non-cancelled list case (this field lives on `EventBaseSchema`, used by list views) incurs zero extra lookups.

## Tests
- `TestEventCancelledNotificationReason::test_built_context_satisfies_schema_and_creates_notification` — real-path regression guard for the crash.
- `TestCanUserSeeCancellationReason` (10 cases) — anonymous / unrelated / invited-only / ticket / cancelled-ticket / RSVP yes / RSVP no / owner / staff / superuser.
- `test_cancellation_reason_visibility.py` — endpoint-level: non-attending authenticated user gets `null`, attending (RSVP yes) sees the reason.

Relevant suite run locally: **89 passed**. `make check` green (ruff, mypy --strict, migrations, i18n, file-length).

## Summary by CodeRabbit

* **New Features**
  * Restricted visibility of event cancellation reasons to attendees and event staff only. Non-attendees will no longer see cancellation reasons.
  * Enhanced cancelled event notifications with refund availability information.

* **Tests**
  * Added tests validating cancellation reason visibility permissions and notification context.

### DIFF
--- a/src/events/models/event.py
+++ b/src/events/models/event.py
@@ -510,6 +510,38 @@ class Event(
 
         return False
 
+    def can_user_see_cancellation_reason(self, user: RevelUser | AnonymousUser) -> bool:
+        """Whether the user may see this event's cancellation reason.
+
+        The reason is addressed to people who were actually going to attend, so
+        it is limited to ticket holders and confirmed RSVPs — plus the event's
+        own staff/owners and superusers. This mirrors the ATTENDEES_ONLY tier of
+        :meth:`can_user_see_address`; a merely-invited or anonymous user does not
+        qualify.
+
+        Args:
+            user: The user to check access for.
+
+        Returns:
+            True if the user may see the cancellation reason, False otherwise.
+        """
+        from .rsvp import EventRSVP
+        from .ticket import Ticket
+
+        if user.is_superuser or user.is_staff:
+            return True
+        if user.is_anonymous:
+            return False
+        if self.organization.owner_id == user.id:
+            return True
+        # Use .all() to leverage prefetched data when available (from with_organization()).
+        if any(m.id == user.id for m in self.organization.staff_members.all()):
+            return True
+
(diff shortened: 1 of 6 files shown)


biagiodistefano:
Good catch @Copilot — refactored in fa27d04. Dropped the bespoke `user_client` fixture and now use the canonical `nonmember_client` / `nonmember_user` fixtures from the controllers `conftest.py`, removing the duplicated user-factory + token-building setup. Tests still green (2 passed), ruff + mypy --strict clean.

Issue #480, opened 2026-06-09
observability: 500 tracebacks are not captured anywhere (no error tracking; django.request errors logged without a stack)


biagiodistefano:
## Summary
When a request 500s, the **traceback is lost**. There is no error-tracking service (no Sentry), and Django's `django.request` "Internal Server Error" log reaches Loki as a bare one-liner with **no stack trace**. This makes production 500s effectively un-debuggable from logs alone.

## Motivating example
A superuser POST to the RevelUser admin changelist 500'd once (1× in 7 days, not reproduced):
- `POST /<admin>/accounts/reveluser/` → `status_code=500`, `user_id=677331de-…`, iPhone Safari, v1.62.1, ~127ms.
- The POST was one of the changelist actions (`send_verification_email` / `send_password_reset_email` / `ban_selected_users`).
- All Loki has for it is: `ERROR  Internal Server Error: /<admin>/accounts/reveluser/` — **no traceback**, so the root cause can't be determined.

## What's missing
1. `django.request` error logs aren't rendered through the structlog `ProcessorFormatter` with `exc_info`, so no stack/exception fields land in Loki.
2. No exception-tracking sink (Sentry/GlitchTip) to capture tracebacks + context for 500s.

## Proposed action
- Ensure unhandled-exception logs include the traceback in structured form (verify the `django.request` logger routes through the JSON `ProcessorFormatter` with `format_exc_info` / `exc_info`), **and/or**
- Add a lightweight error tracker (Sentry or self-hosted GlitchTip) so 500s capture stack + request context.

Without one of these, every production 500 is a dead end (as this reveluser one was).

_The specific reveluser 500 is not separately actionable (one-off, no repro, no stack); filing the observability gap that prevented root-causing it instead._

biagiodistefano:
Update — the actor confirmed the motivating 500 was them **deleting a test user (`test@test.com`) from the Django admin** (matches the `POST /<admin>/accounts/reveluser/` "delete selected" path, time, and user_id).

This makes the point sharper: even *with* the action known, the exact failure can't be determined because **no traceback was captured** (still the core issue here).

Candidate mechanisms in the user-deletion path, for whoever picks this up:
- `RevelUser.delete()` override (`src/accounts/models.py:120-126`) deletes profile-picture thumbnail/preview files from storage before cascading — a storage error there raises.
[...]

Issue #478, opened 2026-06-09
logging: Celery worker logs not emitted as JSON (root-logger hijack adds `[ts: LEVEL/Worker]` prefix)


biagiodistefano:
## Summary
Most of the app logs single-line structlog **JSON**. Celery task logs do **not** — they come out wrapped in Celery's own banner:

```
[2026-06-09 17:00:00,042: INFO/ForkPoolWorker-4] {"event": "final_verification_warnings_sent", "level": "info", ...}
```

i.e. the structlog JSON is prefixed with `[timestamp: LEVEL/Worker]`. This breaks JSON log parsing for the `celery_default` / `beat` services and also produces a misleading stream `level` label (Loki shows some INFO task logs as `warning`).

## Root cause
Celery hijacks the root logger at worker startup (`worker_hijack_root_logger` defaults to **True**) and replaces Django's `structlog.stdlib.ProcessorFormatter` (configured in `src/revel/settings/observability.py`) with its own text formatter `[%(asctime)s: %(levelname)s/%(processName)s] %(message)s`.

- structlog/logging config: `src/revel/settings/observability.py` (JSONRenderer + `ProcessorFormatter` with `foreign_pre_chain`, all handlers use the `json` formatter).
- Celery app: `src/revel/celery.py` — no `setup_logging` signal handler.
- Celery settings: `src/revel/settings/celery.py` — `CELERY_WORKER_HIJACK_ROOT_LOGGER` is not set (so it defaults to True).

## Proposed fix
Let Django own logging. Either:

**Option A (simplest)** — in `src/revel/settings/celery.py`:
```python
CELERY_WORKER_HIJACK_ROOT_LOGGER = False
```

**Option B** — handle the signal in `src/revel/celery.py` so Celery doesn't reconfigure logging:
```python
from celery.signals import setup_logging

@setup_logging.connect
def _keep_django_logging(**kwargs: object) -> None:
    """Keep Django's structlog ProcessorFormatter; don't let Celery hijack logging."""
```

Then verify a worker emits single-line JSON (no `[...]` prefix) and that task context (`task_id`, `task_name`) is still bound.

## Verification
Run a worker and confirm `celery_default` log lines are pure JSON and the `level` is correct.

_Found while building a Loki log-reader CLI: the CLI now strips the `[ts: LEVEL/Worker]` prefix and trusts the embedded JSON's level as a workaround, but the source logs should be JSON like everything else._

biagiodistefano:
[...]

PR #466, opened 2026-05-27
fix(security): harden authz, rate-limiting, and PII/file exposure

biagiodistefano:
Fixes the 7 findings surfaced by the full-codebase security sweep (`/vuln-scan`). All were verifier-confirmed; severities below are post-verification.

## Findings addressed

| Severity | Area | Fix |
|---|---|---|
| HIGH | events-controllers | Organization-admin resource reads (`GET /organization-admin/{slug}/resources` + by-id) now require `IsOrganizationStaff()`. Previously any authenticated user could read another org's PRIVATE/STAFF_ONLY resources and their signed file URLs (read-vs-write authz asymmetry; `for_user` is a visibility filter, not an admin filter). |
| HIGH | questionnaires | `_validate_submission` now asserts each submitted multiple-choice `option_id` belongs to its paired question. The unchecked `(question_id, option_id)` pairing let crafted option IDs (copied from any `is_correct` option) inflate the score and auto-approve admission questionnaires in AUTOMATIC mode. |
| HIGH | settings | `NUM_PROXIES = 1` (env-overridable). With `None`, ninja_extra keyed anonymous throttles on the full client-supplied `X-Forwarded-For`, so a varying header bypassed login/registration/anon throttles behind Caddy (which appends to XFF). |
| HIGH | events-service | `claim_invitation` now rejects hard-blacklisted users on **both** grant paths. `OrganizationStaff` has no status field, so a banned ex-staffer could re-claim a `grants_staff_status` token and regain staff (and bypass `BlacklistGate`, which runs after `PrivilegedAccessGate`). |
| MEDIUM | common | `QuarantinedFile.file` → `ProtectedFileField`, so malware-flagged uploads land under `protected/` and are no longer served from the public `/media/*` Caddy `file_server`. Includes data migration `0014` to relocate existing files (reversible) and an admin `file_link` using `get_file_url()` for authorized forensics downloads. |
| MEDIUM | notifications | EVENT_UPDATED change-diff is now built per-recipient and redacts the changed venue address (old + new) when `can_user_see_address(user)` is False. The dedicated `event_location` field was already gated, but the free-text diff was built once and sent to all recipients. |
| MEDIUM | telegram | `cb_handle_join_waitlist` now resolves the event via `Event.objects.for_user(user)` and runs the full `EligibilityService`, joining only when capacity is the sole obstacle — mirroring the HTTP endpoint. Previously it skipped both, letting ineligible users waitlist (and squat capacity on) events they couldn't see. |

## Notes
[...]

PR #482, opened 2026-06-09
fix: harden questionnaire/poll access control and improve observability

biagiodistefano:
## Summary

Bundles the confirmed findings from a multi-agent vulnerability sweep (`/vuln-scan`) plus one observability follow-up (#478).

### Security

- **Questionnaire answer-key disclosure (HIGH).** `GET /api/questionnaires/{id}` (`get_org_questionnaire`) carried **no permission class** and fell back to the visibility-only `for_user` queryset, so any authenticated user could read the full admission answer key (`is_correct`, weights, `min_score`, `reviewer_notes`, `llm_guidelines`) for every PUBLIC/UNLISTED org's questionnaire — enough to bypass an AUTOMATIC admission gate. Now requires `QuestionnairePermission("evaluate_questionnaire")`, matching its sibling routes.
- **Cross-tenant questionnaire enumeration.** `GET /api/questionnaires/` (`list_org_questionnaires`) had the same missing-permission gap. Now scoped to organizations the caller administers via a new `OrganizationQuestionnaire.objects.for_admin()` (owner/staff only — **no** Django superuser/staff fast-path; platform admins manage questionnaires through the Django admin panel).
- **MC admission-gate bypass (HIGH).** `allow_multiple_answers=False` was enforced only in `MultipleChoiceAnswer.clean()`, but submissions are `bulk_create`d (clean() skipped). Selecting *every* option for a single-answer MC question scored full marks with no penalty and auto-approved AUTOMATIC gates. `_validate_submission` now rejects `>1` option on single-answer questions.
- **DRAFT poll detail disclosure (LOW).** `can_see_poll` ignored `poll.status`, so the detail GET (unfiltered queryset) leaked an unpublished poll's questions/options to anyone knowing the UUID when vote/result visibility was PUBLIC/UNLISTED. Now hidden from non-staff.

### Observability

- **#478** — `CELERY_WORKER_HIJACK_ROOT_LOGGER = False` so Celery worker logs stay single-line structlog JSON instead of being wrapped in Celery's `[ts: LEVEL/Worker]` text banner (which broke JSON parsing and mislabelled the stream level for `celery_default`/`beat`). Task context is bound independently via `structlog.contextvars`, so it's unaffected.

### #480 — investigated, no change needed

The original fix in this branch added a `got_request_exception` receiver to capture admin 500 tracebacks. On review it was **removed as redundant**: Django's uncaught-exception path already logs to `django.request` with `exc_info`, PR #420 added `format_exc_info` to the structlog `FOREIGN_PRE_CHAIN` (so that line renders the full traceback), and `StructlogContextMiddleware` binds the request context (`user_id`/`request_id`/`trace_id`/`path`/`method`) to contextvars that are still active when the 500 is logged. Verified empirically that `django.request` emits the `exception` traceback field under the current config. The receiver only duplicated that line, so #480's traceback requirement is already satisfied by `#420` + the context middleware.

## Notes

- No migrations (queryset methods only, no model field changes).
- The `org_token_staff_ban_bypass` HIGH candidate from the same sweep was **verified already fixed** (PR #466) and needs no change here.
- `ruff format`, `ruff check`, and `mypy` pass on the changed files.

## Test plan

- [ ] `make test` — note: existing questionnaire tests may assert the old unguarded GET/list behavior and need updating alongside the permission/`for_admin` changes.

## Version 1.62.3

* **Bug Fixes**
  * Restricted questionnaire listings to show only those administered by the user
  * Draft polls are now hidden from non-staff users
  * Multiple-choice questions with single-answer requirement now properly enforce this constraint during submission

* **Chores**
  * Version bump to 1.62.3

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "revel"
-version = "1.62.2"
+version = "1.62.3"
 description = "The Revel Event Manager"
 readme = "README.md"
 license = "MIT"
--- a/src/api/apps.py
+++ b/src/api/apps.py
@@ -4,3 +4,7 @@ from django.apps import AppConfig
 class ApiConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "api"
+
+    def ready(self) -> None:
+        """Register the got_request_exception receiver so admin/non-API 500s are logged (#480)."""
+        from api import exception_handlers  # noqa: F401  (import registers the @receiver)
--- a/src/api/apps.py
+++ b/src/api/apps.py
@@ -4,7 +4,3 @@ from django.apps import AppConfig
 class ApiConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "api"
-
-    def ready(self) -> None:
-        """Register the got_request_exception receiver so admin/non-API 500s are logged (#480)."""
-        from api import exception_handlers  # noqa: F401  (import registers the @receiver)
--- a/src/api/exception_handlers.py
+++ b/src/api/exception_handlers.py
@@ -1,5 +1,6 @@
 """Exception handlers for the API."""
 
+import sys
 import traceback
 import typing as t
 from copy import deepcopy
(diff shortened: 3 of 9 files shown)


biagiodistefano:
### Code review

Issue #515, opened 2026-06-15
Show total ticket revenue ("total earned") on the event admin page


biagiodistefano:
## Context
Organizers want accumulated ticket revenue per event for budgeting. Oskar had to filter Stripe manually because Stripe doesn't know which event a payment belongs to.

## Findings (code) — feasible, data exists
- `Payment` stores `amount` (gross), `platform_fee`, `net_amount` (nullable for historical rows), `vat_amount` (`src/events/models/ticket.py:789-797`). Summation precedent: `src/events/service/invoice_service.py:219-233`.
- ⚠️ `net_amount` is nullable on older payments, so a "net" figure may be incomplete.

## Scope
Add a per-event aggregate endpoint. Default the displayed figure to **gross** (+ ticket count); treat net as best-effort.

> 🔗 **Coordination:** frontend dashboard widget counterpart to follow.

---
<sub>Source: Oskar Bechtold feedback session, 2026-06-14. Filed from local triage.</sub>

PR #537, opened 2026-06-16
feat(events): track offline ticket refunds in event revenue (#528)

### LINKED ISSUE #528: Track offline ticket refunds for accurate event revenue (net)
## Context
Follow-up to #515 (per-event revenue endpoint).

When an **offline / at-the-door** ticket is refunded via `mark_offline_ticket_refunded`, the ticket is set to `CANCELLED` and — because offline tickets have **no `Payment` row** — there is no durable marker distinguishing a *refunded* offline ticket from a plain *cancelled* one (`cancel_offline_ticket` produces the same end state).

## Consequence
The #515 revenue endpoint can compute online (Stripe) refunds precisely (`Payment.refund_amount` / `refund_status`), but **offline refunds are invisible**: refunded offline tickets simply drop out of the `gross`/`net` figures. Reported `net` is therefore best-effort for events with offline refunds.

## Possible approaches
- Record the refunded amount on the `Ticket` (e.g. an `offline_refund_amount` field) when `mark_offline_ticket_refunded` runs, and a distinct `cancellation_source`/flag for 'refunded' vs 'cancelled'.
- Or create a lightweight `Payment`-equivalent record for offline collected cash so refunds have a home.

## Acceptance
- Offline refunds are queryable per event and feed into the #515 revenue `refunded`/`net` figures.

cc @bechtold

---
<sub>Filed as a follow-up while implementing #515.</sub>

biagiodistefano:
## Summary

Closes #528 (follow-up to #515, the per-event revenue endpoint).

Offline / at-the-door tickets have **no `Payment` row**, so `mark_offline_ticket_refunded` and `cancel_offline_ticket` produced the same end state (`CANCELLED`). A refunded offline ticket was therefore indistinguishable from a plain cancellation and silently dropped out of `gross`/`net` (the gap was already documented in `get_event_revenue`).

## Changes

- **Model**: new `Ticket.offline_refund_amount` (nullable, `editable=False`) + migration `0080`. Non-null = "this offline ticket was refunded, for this amount"; null = plain cancellation. (Online refunds stay on `Payment.refund_amount`.)
- **Service**: `mark_offline_ticket_refunded` now accepts an optional `refund_amount`:
  - omitted → defaults to the amount paid (`price_paid` or tier price), but **only if the ticket was actually in a paid state** (so refunding a never-checked-in at-the-door ticket records nothing and stays invisible to revenue).
  - explicit → enables **partial refunds**; must be between 0 and the amount paid (else 400).
- **Revenue** (`get_event_revenue`): refunded offline tickets add their **collected** amount back to `gross` and the **refunded** amount to `refunded` — mirroring how the online path keeps a `REFUNDED` payment's full `amount` in gross. Net nets out for full refunds and keeps the remainder for partial ones.
- **Schema**: `refund_amount` added to `AdminCancelTicketSchema`; `offline_refund_amount` exposed on `AdminTicketSchema`.
- **i18n**: new validation string translated for de/fr/it.

## Tests
- Revenue: offline full refund nets to zero; partial refund keeps the remainder; plain cancel stays invisible.
- Refund flow: default records amount paid; explicit partial amount; over-paid amount rejected (400); never-paid at-the-door records nothing.

## Verification
`make check` passes (lint, mypy --strict, migrations, i18n, file-length). The revenue + refund/cancel test suites pass locally (37 tests). Full suite left for CI/review.

## Summary by CodeRabbit

* **New Features**
  * Admins can now record refund amounts when marking offline tickets as refunded, with support for partial refunds and validation to prevent refunds exceeding the paid amount.

* **Bug Fixes**
  * Revenue reports now correctly account for offline ticket refunds in net revenue calculations, matching online refund behavior.


biagiodistefano:
### Code review (self-review, high effort)

8 finder angles + verification. Four findings; the two actionable ones are **fixed in 770a3d2**.

1. **[Medium — correctness, FIXED]** `_resolve_offline_refund_amount` let an *explicit* `refund_amount` through on a ticket that was never paid (e.g. an at-the-door ticket still `ACTIVE`, never checked in), bypassing the `_is_offline_paid` guard the default path enforces. Revenue then added the never-collected tier price to `gross` → fabricated gross/net. Now a non-zero refund on a never-paid ticket is rejected (400) and a zero/omitted one records nothing.
2. **[Low — API clarity, FIXED]** `cancel_ticket` shared `AdminCancelTicketSchema`, so `/cancel` accepted and OpenAPI-documented a `refund_amount` it silently ignored. Split into a dedicated `AdminRefundTicketSchema` used only by `mark-refunded`.
3. **[Low — maintainability, not changed]** `_is_offline_paid` duplicates the `offline_paid` Q in `get_event_revenue` (two sources of truth for 'offline ticket is paid'). Left as-is for now; worth unifying if the paid-state rule grows.
4. **[Low — convention, not changed]** `offline_refund_amount` declared at class level on `AdminTicketSchema` — matches the existing `price_paid`/`discount_amount` pattern on the same schema; cosmetic.

Refuted: a flagged 'TICKET_CANCELLED shows the wrong refund amount' — the cancelled context uses an empty hint, and TICKET_REFUNDED (Payment signal) reads the correct resolved amount.

`make check` green; revenue + refund/cancel suites pass (38 tests, incl. a new never-paid-rejection case).

PR #541, opened 2026-06-17
feat(events): hard-delete unused discount codes, else deactivate (#519)

### LINKED ISSUE #519: Discount code delete: support hard delete when unused, otherwise "deactivate"
## Context
"Delete" on a discount code only soft-deletes (deactivates), so codes look gone when they aren't — the trash icon is misleading.

## Findings (code) — corrects the original premise
- Delete sets `is_active = False` (`src/events/controllers/organization_admin/discount_codes.py:92-103`).
- **There is no Stripe coupon object behind a discount code** — so hard delete is technically possible (the "Stripe constraint" assumption was wrong).
- Usage is knowable: a `times_used` counter exists and `Ticket.objects.filter(discount_code=...)` gives true redemptions.

## Options to discuss
1. Allow **hard delete when unused** (`times_used == 0` and no referencing tickets); otherwise deactivate.
2. Drop hard delete entirely and relabel the action **"Deactivate"** with a non-trash icon.

Either is low effort.

> 🔗 **Coordination:** frontend label/icon counterpart to follow.

---
<sub>Source: Oskar Bechtold feedback session, 2026-06-14. Filed from local triage.</sub>

biagiodistefano:
Closes #519

## Problem
"Delete" on a discount code only soft-deleted it (`is_active = False`), so codes looked gone when they weren't — the trash icon was misleading. It also left the `(organization, code)` slot occupied forever, so an org could never re-create a code string it once "deleted" (e.g. fix a typo'd `SUMMER24`).

## What changed
There is **no Stripe coupon** behind a discount code, so a hard delete is technically safe. The delete endpoint now:

- **Hard-deletes** a code when it was never used (`times_used == 0` **and** no referencing tickets) — this also frees the `(organization, code)` slot for reuse.
- **Deactivates** (`is_active = False`) a used code, preserving the redemption link on its tickets.

The dual guard is belt-and-suspenders: `times_used` never decrements, but a cancelled-ticket scenario could leave `times_used == 0` while tickets still reference the code. Since `Ticket.discount_code` is `on_delete=SET_NULL`, hard-deleting a used code would null out historical attribution — so we don't.

## API change ⚠️
The endpoint switches from `204 No Content` to `200` with a body so the frontend knows which action occurred:

```json
{ "action": "deleted" | "deactivated" }
```

- Requires a regenerated TS client (`pnpm generate:api`).
- **Frontend counterpart** (label/icon) tracked separately per the issue.

## Files
| File | Change |
|------|--------|
| `events/service/discount_code_service.py` | New `delete_discount_code(dc)` returning `"deleted"`/`"deactivated"`. |
| `events/schema/discount_code.py` | New `DiscountCodeDeleteResponse` (+ re-exported from `schema/__init__.py`). |
| `events/controllers/organization_admin/discount_codes.py` | `delete_discount_code` returns `200` + body, delegates to service (thin controller). |
| tests (controller + service) | Rewrote old soft-delete tests; added both branches incl. the "referenced by ticket but `times_used == 0`" edge. |

## Verification
- `make format`, `make lint`, `make mypy` all pass.
- Tests cover hard-delete, deactivate-via-counter, and deactivate-via-referencing-ticket branches.

## Release Notes

* **New Features**
  * Discount code deletion now returns a response indicating the action taken (deleted or deactivated)
  * Unused discount codes are now permanently removed; codes with usage history are deactivated instead
  * HTTP response status for deletion changed from 204 to 200

### DIFF
--- a/src/events/controllers/organization_admin/discount_codes.py
+++ b/src/events/controllers/organization_admin/discount_codes.py
@@ -92,12 +92,11 @@ class OrganizationAdminDiscountCodesController(OrganizationAdminBaseController):
     @route.delete(
         "/discount-codes/{code_id}",
         url_name="delete_discount_code",
-        response={204: None},
+        response={200: schema.DiscountCodeDeleteResponse},
     )
-    def delete_discount_code(self, slug: str, code_id: UUID) -> tuple[int, None]:
-        """Soft-delete a discount code by deactivating it."""
+    def delete_discount_code(self, slug: str, code_id: UUID) -> schema.DiscountCodeDeleteResponse:
+        """Delete a discount code: hard-delete when unused, otherwise deactivate."""
         organization = self.get_one(slug)
         dc = get_object_or_404(models.DiscountCode, pk=code_id, organization=organization)
-        dc.is_active = False
-        dc.save(update_fields=["is_active"])
-        return 204, None
+        action = discount_code_service.delete_discount_code(dc)
+        return schema.DiscountCodeDeleteResponse(action=action)
--- a/src/events/schema/__init__.py
+++ b/src/events/schema/__init__.py
@@ -254,6 +254,7 @@ from .follow import (
 # Discount code schemas
 from .discount_code import (
     DiscountCodeCreateSchema,
+    DiscountCodeDeleteResponse,
     DiscountCodeSchema,
     DiscountCodeUpdateSchema,
     DiscountCodeValidationResponse,
@@ -531,6 +532,7 @@ __all__ = [
     "OrganizationFollowUpdateSchema",
     # Discount codes
     "DiscountCodeCreateSchema",
+    "DiscountCodeDeleteResponse",
     "DiscountCodeSchema",
     "DiscountCodeUpdateSchema",
     "DiscountCodeValidationResponse",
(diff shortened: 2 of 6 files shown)


biagiodistefano:
Frontend counterpart (API contract change + trash-icon relabel) tracked in letsrevel/revel-frontend#466.

biagiodistefano:
### Code review

Found 1 issue:

1. `delete_discount_code` has no transaction boundary or row lock, so the check-then-delete is racy. Between `dc.tickets.exists()` and `dc.delete()` (and because `dc.times_used` is read from an instance fetched without a lock), a concurrent purchase can `bulk_create` a ticket referencing the code and increment `times_used`. Since `Ticket.discount_code` is `on_delete=SET_NULL`, the hard-delete then nulls the FK on a just-paid ticket, losing redemption history (and the discount attribution that `attendee_invoice_service` reads). The sibling `apply_discount` in this same file deliberately guards this with `DiscountCode.objects.select_for_update().get(pk=...)` inside `@transaction.atomic`; this new function should mirror that. (revel-backend CLAUDE.md: "For create operations with uniqueness constraints, use `get_or_create_with_race_protection()`" / race-condition protection, and the `select_for_update` precedent already in this module.)

https://github.com/letsrevel/revel-backend/blob/c7bb10be5f341e8a1c6c74eab5e19ca377fb6487/src/events/service/discount_code_service.py#L151-L157

Issue #552, opened 2026-06-19
feat(events): scheduled (quarterly/monthly) revenue report delivery — depends on #551


biagiodistefano:
## Summary

Add **opt-in scheduled delivery** of the revenue & VAT report bundle (from #551) to organizers,
aligned to the Austrian UVA filing cadence. A per-org setting picks the cadence; a Celery beat
job generates the just-closed period's report and emails the ZIP.

**Depends on #551** (the on-demand report bundle). This issue is purely the scheduling/delivery
layer on top of it.

## Motivation

The on-demand report (#551) covers "download whenever". A proactive periodic email is a cheap
retention/professionalism win: it nudges orgs toward their UVA deadline and gives their accountant
a predictable document. The cadence should match the org's actual VAT rhythm:

- **Monthly UVA** — only if prior-year turnover > €100,000
- **Quarterly UVA** — turnover €55,000–€100,000 (due 15 May / 15 Aug / 15 Nov / 15 Feb)
- **Under €55,000** — generally no periodic UVA, just the annual return

Most Revel orgs (community/cultural) are under €100k → **quarterly is the right default**; monthly
is for the high-turnover minority. Revel can't infer an org's total turnover, so the org (or its
accountant) chooses.

## Design

- **Per-org setting** `revenue_report_cadence`: `NONE` (default) / `QUARTERLY` / `MONTHLY`.
  Requires `billing_email` to be set before enabling (reuse invoicing-prerequisite style validation).
- **Celery beat job** (`send_scheduled_revenue_reports_task`), run early each month:
  - For each org with a cadence whose period just closed (calendar quarter for `QUARTERLY`,
    calendar month for `MONTHLY`), call the existing
    `get_or_generate_revenue_report(org, scope=closed_period)` from #551.
  - **Settle window:** generate ~2 weeks after period close so late refunds/chargebacks settle,
    still leaving ~a month before the 15th-of-second-month deadline.
  - **Skip empty periods** (no sales → no email).
  - Email the ZIP to `billing_email` (+ owner), reusing the invoice-email delivery path.
- **Snapshot caveat in the email body:** "Snapshot as of {date}. If refunds/chargebacks occur
  later, re-download the current version from your dashboard." The on-demand report (#551) remains
  the source of truth; the emailed bundle is a point-in-time nudge.

## Cadence period boundaries

[...]

Issue #551, opened 2026-06-19
feat(events): downloadable revenue & VAT report bundle (XLSX + PDF in ZIP)


biagiodistefano:
## Summary

Add an on-demand, cached **revenue & VAT report** for organizers — a downloadable
**ZIP bundle containing an XLSX (detail + summary) and a PDF (summary cover)** — that an
organizer can hand to their accountant for the Einnahmen-Ausgaben-Rechnung / UVA.

This is an **internal accounting document, not an invoice**. It reports the **output
(revenue) side only**. It deliberately does **not** carry a sequential invoice number
(would collide with `AttendeeInvoice`) and does **not** list Stripe or Revel platform
fees (see "Why no fees").

## Motivation

`get_event_revenue()` (`events/service/ticket_service.py`) returns gross/refunded/net per
currency, but accountants need **output VAT split per rate** (UVA has separate boxes for
20% / 13% / 10% / 0%-reverse-charge) plus a transaction-level detail to reconcile against
the bank/Stripe payouts. There is no such artifact today, and the ticket PDF carries no
financial fields. Per-attendee invoices are **not legally required** for B2C ticket sales
in AT, so a per-period summary is the right unit.

## Why no fees (avoids double-booking)

- **Stripe** issues its own monthly fee invoice/document (Dashboard → Documents). For an
  AT business the Stripe fee is an intra-EU service from Stripe Ireland → **reverse-charge**
  handled by the accountant from that document.
- **Revel** issues `PlatformFeeInvoice` to organizers separately.

If this report also listed those as cost lines, they'd be booked twice. So fees are
excluded; a footnote points the reader to the source invoices.

## Scope of the report (output side only)

- Gross ticket revenue
- − Refunds / credit notes (in the period they occurred, as negatives)
- = **Net taxable turnover**
- **Output VAT split by rate** (incl. a 0% / reverse-charge bucket for B2B & non-EU,
  reusing `attendee_vat_service`)
- Ticket counts (sold / refunded)
- Per **currency** section (data is multi-currency; see `get_event_revenue`)
- Footnote: *"Bank payouts are net of Stripe processing fees and Revel platform fees,
  which are booked separately from their respective monthly invoices."*

## Bundle contents

**PDF** (`render_pdf` + new template `templates/reports/revenue_vat_report.html`):
- Seller identity (org `billing_name`, `vat_id`, `billing_address`)
- Scope (date range, or single event), currency, generated-at, source = Stripe
[...]

Issue #550, opened 2026-06-18
seed: REFUNDED payments lack refund_amount/refund_status (ticket stays ACTIVE) → revenue net overstated, "0 paid tickets"


biagiodistefano:
## Summary

Seeded **`REFUNDED`** payments are created without a recorded refund (`refund_amount` / `refund_status`) and without cancelling the ticket. This makes `get_event_revenue()` **overstate `net`** and report a self-contradictory figure to the new FE "Total earned" card (letsrevel/revel-frontend#451): e.g. **`$45.00` earned but `0 paid tickets`**, with **no** refund shown (because `net == gross`).

## How it manifests

The FE event-ticket-admin revenue card renders the `/api/event-admin/{event_id}/revenue` aggregate. On seeded data it shows a gross amount with `0 paid tickets` and no net/refunded line — implying money was earned but no tickets are held and nothing was refunded, which is impossible for a real Stripe refund.

## Root cause

`src/events/management/commands/seeder/tickets.py` (the `seed` command — `make seed`) assigns a payment status from `payment_status_weights` (`seeder/config.py` → `"refunded": 0.05`, ~5% of online payments). When the draw is `"refunded"` it builds:

```python
Payment(
    ...,
    status=Payment.PaymentStatus.REFUNDED,
    amount=amount,
    ...
)
```

but **never sets `refund_amount` or `refund_status=SUCCEEDED`**, and **leaves the ticket `ACTIVE`** (no `CANCELLED`, no `offline_refund_amount`). The real Stripe refund flow records the refund and transitions the ticket; the seed does not, so the row is in a state production never produces.

### Why the aggregate then misreports

`get_event_revenue()` (`src/events/service/ticket_service.py:895`) computes, for online payments:

- `gross` = `Sum(amount)` over `status in [SUCCEEDED, REFUNDED]`  → **counts the refunded $45**
- `refunded` = `Sum(refund_amount)` over `refund_status == SUCCEEDED` → **0** (never set)
- `paid_ticket_count` = `Count(status == SUCCEEDED)` → **0** (it's `REFUNDED`)
- `net = gross - refunded` → **$45** (overstated; should be $0)

The service is **correct given consistent data** — this is purely a seed-data defect.

## Scope note

[...]

PR #554, opened 2026-06-19
feat(events): revenue & VAT report bundle, scheduled delivery + org financial endpoints (#550, #551, #552, #556)

biagiodistefano:
Lands four related changes for organizer revenue reporting as one PR.

## #550 — seed refund consistency (fix)
Seeded `REFUNDED` payments now mirror the production refund path: they record
`refund_amount`/`refund_status=SUCCEEDED`/`refunded_at`/`stripe_refund_id`, cancel
the ticket (`CANCELLED`, `STRIPE_DASHBOARD`), and decrement `tier.quantity_sold`.
Mix of ~70% full / ~30% partial refunds.

## #551 — on-demand revenue & VAT report bundle (feature)
New revenue report service:
- Aggregates **full turnover** (online `Payment` rows + offline/at-the-door tickets)
  per **currency** and per **VAT rate**, plus a per-transaction detail. Refunds are
  attributed to the period they occurred, as negatives. VAT uses the stored
  `net/vat/rate` when present, else derives from `org.vat_rate`; reverse-charge /
  non-EU land in the 0% bucket.
- Builds a **ZIP** bundling an **XLSX** (Summary + Transactions sheets) and a **PDF**
  (per-rate VAT table, refunds, net taxable turnover, "internal document — not an
  invoice" banner + fee footnote).
- Cached via `FileExport` keyed on a SHA-256 `data_hash` (over in-scope payment/ticket
  rows **+ org VAT config**), with `?refresh=true` override.
- Org-admin endpoints under `/organization-admin/{slug}`: `POST /revenue-report`
  (generate/reuse) and `GET /revenue-reports/{id}` (poll + signed URL), guarded by
  `OrganizationPermission("manage_organization")`.

## #552 — scheduled delivery (feature)
- Per-org `revenue_report_cadence` (NONE/QUARTERLY/MONTHLY); enabling requires
  `billing_email`.
- `get_organization_timezone` (org city tz → platform default) + `closed_period_for`
  (quarter/month boundaries, Q4→Jan rollover).
- Monthly Celery beat job (06:00 on the 16th, Europe/Vienna) generates the just-closed
  period's report and emails the ZIP to `billing_email` + owner; skips empty periods;
  idempotent via a per-period marker; per-org failures are isolated.

## #556 — org financial endpoints + unified engine (feature, addendum)
A live, on-screen financial view to pair with the downloadable report.
- **Unifies the two divergent revenue engines into one tax-precise engine**
  (`events/service/revenue_aggregation.py`): a single per-row pass accumulates at
  `(event, currency, vat_rate)` grain. The downloadable report (org roll-up), the new
  org endpoint (group-by-event), and the per-event endpoint (filter to one event) are
  all projections of it — so every view reconciles. The shipped report's XLSX/PDF/ZIP
  output, cache, and scheduled delivery are unchanged (parity-gated by the existing
  report tests through the Task 1-3 refactor).
- **New endpoint** `GET /organization-admin/{slug}/revenue` → per-event financials:
  sortable (`sort=revenue|event_start`, `order`), period-filtered (`year` + optional
  `month`/`quarter`, mutually exclusive → 422; default current year in org tz),
  dominant-currency default + optional `?currency=` filter (`available_currencies`
  returned for a switcher). Permission `manage_organization`.

### ⚠️ Breaking change (per-event endpoint)
`GET /event-admin/{id}/tickets/revenue` is re-backed onto the unified engine and now
returns `EventFinancialsSchema` (all-time default). Field changes:
- `refunded` → `refunds`
- `paid_ticket_count` **removed** — derive as `sold_count - refunded_count`.
  **`sold_count` is not the old number**: it counts later-refunded/cancelled sales too.
- VAT detail added: `net_taxable`, `vat`, `rate_buckets`.

Requires a coordinated frontend `pnpm generate:api` + consumer update (tracked in #556).

## Notes
- Migrations: `common 0015` (export type), `events 0084` (org fields), `events 0085`
  (beat schedule). No new migrations for #556 (read-only aggregation).
- Revenue Celery tasks live in `events/revenue_tasks.py`; the aggregation engine lives in
  `events/service/revenue_aggregation.py` (keeps both service files under the 1000-line limit).
- Follow-ups (non-blocking): `TxnRow.date` shows the sale date for cross-period
  refund-only rows (add a `refund_date` field); scheduled generation runs synchronously
  in the beat worker (fine at current scale; fan-out if org count grows).

## Post-review hardening (#556)
Follow-up commit addressing findings surfaced after the initial review:
- **Tests no longer time-fragile**: the revenue aggregation/bundle/cache tests pinned a
  hardcoded 2026 window against `now`-stamped rows and would have broken after 2026 —
  widened to a year-agnostic scope.
- **Restored per-event boundary guards** dropped in the endpoint rewrite: at-the-door
  `ACTIVE` excluded (counts only once `CHECKED_IN`), pending-only event returns empty,
  offline partial-refund keeps the remainder in net, and PWYC `price_paid` override.
- **Input validation**: `POST /revenue-report` now rejects an inverted date range
  (`date_from > date_to`) with `422` instead of silently returning an empty report.
- **Type safety**: `_parameters_to_scope` is typed with the `_ScopeParameters` TypedDict
  (no raw dicts); clarified pre-refund vs net-of-refunds in the financials docstrings.

## Testing
Full `make check` green (ruff, strict mypy across 834 files, migration-check, i18n,
file-length). `make test`: **5315 passed, 15 skipped**, total coverage **96%**.


biagiodistefano:
## Review resolution (commit f32c6fd)

Went through the CodeRabbit comments and an internal multi-agent review. Summary of what changed and what was intentionally left.

**Addressed**
- **`event_id` validation** (`src/events/schema/revenue_report.py`, `src/events/controllers/organization_admin/revenue.py`): typed `event_id` as `UUID | None` so a malformed value returns 422 at the schema boundary instead of a 500 from `UUID(...)` in the controller. Added `test_malformed_event_id_returns_422`.
- **Typed scope parameters** (`src/events/service/revenue_report_service.py`): `_scope_to_parameters` now returns a `TypedDict` (CLAUDE.md: "avoid raw dicts").
- **Offline 0-refund count** (`revenue_report_service.py`): cancelled offline tickets with a `0` refund now increment `refunded_count` (guard changed to `is not None`). Added a regression test.
- **Org-local email snapshot date** (`revenue_report_service.py`): the scheduled-report email "Snapshot as of" date now uses the org's local date, not UTC.

**Deferred (tracked follow-ups, non-blocking)**
- **`TxnRow.date` for cross-period refund-only rows**: the Transactions sheet shows the sale date for a refund whose sale fell outside the period. Summary totals/counts are unaffected (they use the refund date for period membership). Documented in-code with a `# ponytail:` note; the fix is a `refund_date` field on `TxnRow`.
- **`compute_revenue_data_hash` runs before the cache lookup**: the hash is also needed on a miss (it is stored on the new export's `parameters`), so deferring it past the lookup saves nothing on the hot path. A real fix means a cheaper DB-aggregate invalidation signal, which trades cache-correctness safety for speed — left for a dedicated change.

**Declined (with reasoning)**
- **Validate cadence against the effective billing email** (`organization_service.py`): `billing_email` is not a field on `OrganizationEditSchema`, so it cannot be set in the same payload as `revenue_report_cadence`; checking the persisted value is correct here.
- **Materialize org IDs + `select_for_update` in the scheduled sweep**: the loop uses a plain queryset (no `.iterator()` server-side cursor), so the engineering-notes anti-pattern does not apply. Holding a row lock across PDF generation would be a worse pattern, and an optimistic claim-before-send would break the deliberate retry-on-failure behavior. Left as-is for a monthly, low-org-count beat job.
- **Test polish** (include `FAILED` in poll statuses; extract aggregation fixtures to `conftest`): minor test-quality suggestions, not adopting in this PR.

`make check` is green (ruff, strict mypy across 828 files, migration-check, i18n, file-length); revenue-feature tests pass.

biagiodistefano:
### Code review

Issue #608, opened 2026-06-23
refactor: consolidate the 8 account email-send wrappers into one template-driven task


biagiodistefano:
Follow-up to #607 (the `accounts/tasks.py` split shipped separately). This tracks the **behavior-changing** half that was deliberately kept out of the pure-move PR.

## Summary

`accounts/tasks/email.py` contains **8 near-identical thin email-send wrappers**, each of which renders a subject + txt/html body and calls `send_email`:

- `send_verification_email`
- `send_account_activation_link`
- `send_password_reset_link`
- `send_email_change_confirmation`
- `send_email_change_notice`
- `send_email_change_completed_old`
- `send_email_change_completed_new`
- `send_account_deletion_link`

They share the same shape but differ in: template base name, the frontend link path built from a token, the context keys, and the recipient-argument name. Consolidating them into a single template-driven task removes most of the duplication.

## Why this is its own issue (not part of the split)

Unlike the split (a pure move, #607), this **changes task signatures and every dispatch call site**, so it needs design discussion + focused tests:

- A generic `send_account_email(*, to, template_base, context, ...)` (or a small config table keyed by an enum) has to faithfully reproduce each wrapper's link path and context.
- Call sites currently do `from accounts.tasks import send_verification_email; ....delay(email, token)` — each must migrate to the consolidated signature.
- The registered Celery task name(s) change. Either keep thin pinned-name shims that delegate to the generic task (zero risk, less dedup) or migrate dispatchers + accept new names (more dedup, needs an in-flight-message deploy plan).

## Acceptance criteria

- [ ] The 8 wrappers consolidated into a template-driven task (or a minimal shim + generic core)
- [ ] All dispatch call sites updated; email contents (subject, links, context) byte-for-byte equivalent — assert via rendered output in tests
- [ ] Celery task-name strategy decided and documented (shims vs. migration); no beat/migration impact (none of these 8 are beat-scheduled)
- [ ] `make check` green; email tests cover each message type

## Watch item (carried from #607, not in scope here)

`questionnaires/service/questionnaire_service.py` is **966 lines** — extract a `SubmissionService` before it crosses the 1000-line gate.

---
_Split out from #607. Source: tech-debt assessment 2026-06-23 (v1.65.0), Medium #6._

PR #609, opened 2026-06-23
refactor(tasks): package the accounts/events task modules + enforce explicit Celery task names (#607)

### LINKED ISSUE #607: refactor: split accounts/tasks.py + consolidate email-send wrappers
## Summary

Split-out follow-up to #575. The `events/tasks.py` half shipped in #606; this issue tracks the **`accounts/tasks.py`** half, which is deliberately separated because it carries extra risk the events half did not.

`src/accounts/tasks.py` is **995 lines** — one feature away from the hard 1000-line CI file-length gate.

## Why this half is riskier than events (must read before splitting)

Unlike `events/tasks.py` (where every beat-scheduled task already had an explicit `name="events.X"`), **six `accounts` beat-scheduled tasks rely on the *default* module-path name** `accounts.tasks.<func>`, referenced **by string** in periodic-task migrations:

- `accounts.tasks.cleanup_expired_data_exports`
- `accounts.tasks.deactivate_unverified_accounts`
- `accounts.tasks.delete_old_inactive_accounts`
- `accounts.tasks.flush_expired_tokens`
- `accounts.tasks.send_early_verification_reminders`
- `accounts.tasks.send_final_verification_warnings`

(`accounts.process_referral_payouts` already has an explicit name and is safe.)

Moving any of those six to a submodule changes its registered Celery name from `accounts.tasks.<func>` to `accounts.tasks.<submodule>.<func>`, which **silently breaks the beat schedule** (the `PeriodicTask` row still references the old string → worker raises `NotRegistered` → the scheduled job stops firing).

**Verified empirically** (sweep in #606): forcing full task discovery, all 28 beat task names currently resolve; the six above resolve only because they still live in `accounts.tasks`.

### Required mitigation
Pin `name="accounts.tasks.<func>"` (the current default) on every bare task that moves, so the registered name stays byte-identical — exactly the pattern used for the 9 bare events tasks in #606. Then verify with a registry cross-check + `makemigrations --check`.

## Suggested splits (from the audit)

- `tasks_email.py` — email lifecycle wrappers
- `tasks_gdpr.py` — GDPR / cleanup
- `tasks_payouts.py` — referral payouts

Keep public import paths working via re-export from `accounts/tasks/__init__.py` (package), mirroring #606.

## Bundled cleanup (audit rec #6)

`accounts/tasks.py` contains **8 near-identical thin email-send wrappers** (verification, password reset, email change ×N, deletion). Consolidating them into a single template-driven task is a natural part of the email-cluster split and removes most of the length pressure on its own. Unlike #606 (a pure move), this is a behavior-affecting change — keep it in its own commit with focused tests.

## Watch item (next on the list, not in scope here)

`questionnaires/service/questionnaire_service.py` is **966 lines** — extract a `SubmissionService` before it crosses 1000.

## Acceptance criteria

- [ ] `accounts/tasks.py` split into focused modules, comfortably under 1000 lines
- [ ] All Celery task names unchanged — the six default-named beat tasks pinned to `accounts.tasks.<func>`; verified against periodic-task migrations + live registry
- [ ] Email-send wrappers consolidated (separate commit, with tests)
- [ ] `make check` passes (file-length + mypy + ruff); affected account-task tests pass

---
_Split out from #575 (events half shipped in #606). Source: tech-debt assessment 2026-06-23 (v1.65.0), High-Priority #1 + Medium #6._

biagiodistefano:
Closes #607. Three related Celery-task changes, all behavior-preserving, folded into one PR (disjoint files). The email-wrapper consolidation (rec #6) remains tracked separately in **#608**.

---

## 1. Split `accounts/tasks.py` into a per-domain package (#607)
`accounts/tasks.py` was **995 lines**. Split into an `accounts/tasks/` package: `email`, `tokens`, `gdpr`, `verification_reminders`, `notifications`, `bans`, `payouts`.

**Careful bits:** six accounts beat tasks rely on the *default* `accounts.tasks.<func>` name (referenced by string in migrations), so all 19 bare tasks are **pinned** to `name="accounts.tasks.<func>"` (byte-identical); `process_referral_payouts` keeps `accounts.process_referral_payouts`. `mark_reminder_sent` is re-exported (the email-callback allowlist resolves it via `getattr(import_module("accounts.tasks"), …)`). `__init__.py` re-exports every task + `DATA_EXPORT_URL_EXPIRES_IN`. Tests retarget three module-global patches into their submodules.

## 2. Move stray events task modules into the `events/tasks/` package
`events/tasks_stripe.py` → `tasks/stripe_webhooks.py`, `events/announcement_tasks.py` → `tasks/announcements.py`, `events/revenue_tasks.py` → `tasks/revenue.py` (`git mv`, history preserved). Inside the package, autodiscovery registers them, so the manual `EventsConfig.ready` import is dropped. All keep their explicit `events.*` names → beat schedules in migrations 0079/0082/0085 unaffected.

## 3. Enforce explicit Celery task names + orphan detection (new)
The splits above surfaced the root footgun — a bare `@shared_task` takes Celery's default `<module>.<func>` name, so moving a module silently renames the task and breaks beat. Made it impossible to reintroduce:

- **Static check** `scripts/check_task_names.py` — AST-based (no imports/DB), fails if any `@shared_task`/`@app.task` lacks `name=`. Wired into **`make check`** (new `task-names` target) **and CI** (`test.yaml` step), mirroring the file-length gate.
- **Pinned the 17 previously-bare tasks** across `common`, `common.thumbnails`, `geo`, `notifications`, `questionnaires` — each to its **current default name**, so every registered name (incl. beat-scheduled ones) stays byte-identical.
- **Runtime check** `manage.py check_orphaned_beat_tasks` — fails if any `PeriodicTask` row points at an unregistered name (catches manually-created/admin rows the static check can't see); suitable as a deploy/health gate. Both paths verified (happy + injected orphan → exit 1).
- **Docs**: `CLAUDE.md` "Other conventions" gains the pin-`name=` rule + both checks; new "Celery Task Names" section in `docs/engineering-notes.md` explains the failure mode.

---

## Migration safety (swept across all three changes + #606)
- **No migration imports any moved/renamed/split task module by Python path**, and none use `import_string`/`importlib`/`__import__` — nothing breaks at graph-load or `RunPython` execution.
- Every migration→task link is a **name string** in a `PeriodicTask` row; all preserved (splits/pins keep names byte-identical; the move changed only file paths).
- `make check`: no missing migrations.

## Validation
- `make check` green: ruff, `mypy --strict` (866 files), migrations, i18n, file-length, **task-names**.
- Runtime: every task registers under its original name (no drift); full **28-name beat sweep resolves (UNRESOLVED: NONE)**; `mark_reminder_sent` callback + `check_orphaned_beat_tasks` (both paths) verified.
- ~560 affected tests pass locally (accounts/events task & service + the 5 name-pinned apps).

## Out of scope
Email-wrapper consolidation (#608).


biagiodistefano:
### Code review

PR #610, opened 2026-06-23
refactor(events): move standalone task modules into the events/tasks package

biagiodistefano:
Harmonization follow-up to the task-package splits (#606/#607). Three events Celery-task modules sat *beside* the `events/tasks/` package — leftovers from when `events/tasks.py` was at the 1000-line gate and they needed manual registration via `EventsConfig.ready`. Now that `events/tasks` is a package, they belong inside it.

## What
| from | to |
|---|---|
| `events/tasks_stripe.py` | `events/tasks/stripe_webhooks.py` |
| `events/announcement_tasks.py` | `events/tasks/announcements.py` |
| `events/revenue_tasks.py` | `events/tasks/revenue.py` |

(`git mv`, so history is preserved.)

- **`EventsConfig.ready` simplified.** It imported these three modules purely so the worker would register their tasks (autodiscovery doesn't find modules not named `tasks`). Inside the package, Celery autodiscovery registers them like every other submodule, so the manual import is removed — `signals` stays. Verified at runtime: all five tasks register via autodiscovery under their original names.
- **`__init__.py` re-exports** the five tasks; the three external import sites now use the package path; the `announcements` relative import `.models` → `events.models`. Stale docstrings (referencing the old location / `ready()` wiring) updated.

## Behavior-preserving
All tasks keep their explicit registered names (`events.prune_stripe_webhook_events`, `events.send_scheduled_announcements`, `events.resend_announcements_to_new_signups`, `events.generate_revenue_report`, `events.send_scheduled_revenue_reports`), so the Celery-beat schedules in migrations **0079 / 0082 / 0085** — which reference tasks by name string — are unaffected.

## Migration safety (swept across this rename + #606 + #607)
- **No migration imports any moved/renamed/split task module by Python path**, and none use `import_string`/`importlib`/`__import__` — so nothing breaks at migration graph-load or `RunPython` execution.
- Every migration→task link is a **name string** in a `PeriodicTask` row; all such names are preserved (rename changed only file paths; #607's six default-named `accounts.tasks.*` were name-pinned).
- `make check`: **no missing migrations** (moving `.py` files doesn't alter the graph).

## Validation
- Runtime: all 5 moved tasks register via autodiscovery; full 28-name beat sweep resolves (UNRESOLVED: NONE).
- `make check` green: ruff, `mypy --strict` (859 files), migrations, i18n, file-length.
- 51 affected task/service tests pass locally.

### DIFF
--- a/src/events/apps.py
+++ b/src/events/apps.py
@@ -8,15 +8,14 @@ class EventsConfig(AppConfig):
     def ready(self) -> None:
         """Run startup-time side effects for the events app.
 
-        Imports the signal handlers and the ``tasks_stripe`` /
-        ``announcement_tasks`` / ``revenue_tasks`` modules (so the Celery worker
-        registers their tasks) and installs the per-app exception handlers on the
-        global Ninja API.
+        Imports the signal handlers and installs the per-app exception handlers on
+        the global Ninja API. The Celery tasks live in the ``events.tasks`` package
+        and are registered via Celery autodiscovery, so they need no manual import here.
 
         Returns:
             None: Performs registration side effects only.
         """
-        from . import announcement_tasks, revenue_tasks, signals, tasks_stripe  # noqa: F401
+        from . import signals  # noqa: F401
         from .exception_handlers import register as register_exception_handlers
 
         register_exception_handlers()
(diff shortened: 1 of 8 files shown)


biagiodistefano:
Folded into #609 — the events task-module move now rides in the same PR as the accounts/tasks split (disjoint files, single review). Closing this in favor of #609.

PR #617, opened 2026-06-24
fix(billing): track email delivery so financial documents aren't silently lost (#616)

### LINKED ISSUE #616: fix(billing): track email delivery so a statement/invoice isn't silently lost when SMTP retries exhaust
## Summary

Both `accounts.generate_and_send_payout_statement` (added in #615) and `events.deliver_attendee_invoice` follow the same shape: a retryable Celery task that (1) generates/loads a financial-document PDF, then (2) emails it. Generation is idempotent, but **email delivery is not tracked anywhere**. If the send step fails on every attempt (initial + `max_retries`), the document exists but the recipient never receives it — with **no recovery path**.

For payout statements specifically, #615's backstop sweep (`_redispatch_missing_statements`) only re-dispatches payouts where `statement__isnull=True`. Once the statement row exists, the sweep skips the payout forever — so a delivery that exhausted its retries is silently dropped.

## Why it matters

These are financial documents (self-billing Gutschrift / payout statement / attendee invoice). Silently never delivering one is more than cosmetic.

## Where

- `src/accounts/tasks/payouts.py` — `generate_and_send_payout_statement` (statement) + `_redispatch_missing_statements` (sweep query keyed on `statement__isnull=True`).
- `src/events/tasks/invoicing.py` — `deliver_attendee_invoice_task` / `deliver_attendee_credit_note_task` share the same untracked-delivery property.

## Suggested approach

Persist a delivery state on the statement/invoice (e.g. `email_sent_at` / a small outbox status), set it only after a successful `send_email`, and make the recovery sweep select `PAID` payouts (and issued invoices) missing **either** the document **or** a successful delivery. Apply it consistently across both invoice and statement delivery rather than piecemeal — ideally a shared helper/mixin.

Note: re-sending on a retry already produces at-least-once email semantics today; this issue is about the **at-most-zero** failure mode (delivery never succeeds and is never retried again), not deduplicating duplicates.

## Context

Surfaced by CodeRabbit during review of #615 (https://github.com/letsrevel/revel-backend/pull/615). Deliberately kept out of #615 to keep that fix surgical and because it needs a model change + should cover both delivery paths.

biagiodistefano:
## Summary

Financial-document delivery was never tracked: if an email send exhausted its Celery retries, the document existed but the recipient never received it — with no recovery path. The payout backstop sweep only re-selected payouts with no statement (`statement__isnull=True`); the attendee invoice/credit-note and platform-fee paths had **no** recovery sweep at all.

Closes #616.

## What changed

- **`EmailDeliverableMixin`** (`common/models/base.py`) — adds `email_sent_at` + an idempotent `mark_email_sent()` (first-write-wins, so a re-send never moves the recorded time). Mixed into `ReferralPayoutStatement`, `AttendeeInvoice`, `AttendeeInvoiceCreditNote`, `PlatformFeeInvoice`.
- **Mark only after a successful send:**
  - Payout statement: marks after the existing synchronous send.
  - Attendee invoice/credit note: switched `_send_org_branded_email` from `.delay()` to a **synchronous** send so the already-retryable deliver task owns retries and marks on success. Both deliver functions now **self-heal a missing PDF** (`ensure_pdf_exists` / `ensure_credit_note_pdf_exists`), so a single re-dispatch recovers a document missing *either* its PDF *or* its delivery.
  - Platform-fee invoice: added `ensure_invoice_pdf_exists` (the PDF is saved outside the create transaction, so an ISSUED invoice can lack one) + marks after send.
- **Recovery sweeps:**
  - Widened `_redispatch_missing_statements` to `statement IS NULL OR statement.email_sent_at IS NULL`.
  - New `events.redispatch_undelivered_invoices` beat task (every 15 min, data migration `0089`) covering attendee invoices, credit notes, and platform-fee invoices.
- **Edge case:** a document with no recipient is a permanent failure — not marked, harmlessly re-swept (documented inline).

## Migrations
- `accounts/0028`, `events/0088` — add `email_sent_at`.
- `events/0089` — create the 15-min `PeriodicTask` row.

## Testing
- `make check`: ✅ format, lint, mypy --strict, migration-check, i18n, file-length, task-names.
- Ran the affected test files: **158 passed**. Updated the sweep/delivery tests broken by the wider query + sync switch, and added tests for delivery marking, idempotency, PDF self-heal, and the new sweep across all four document types.

## Note
The attendee delivery path is now synchronous, so an SMTP outage blocks the deliver worker for the send duration rather than returning immediately — acceptable since each document is retry-isolated in its own task.

## Summary by CodeRabbit

* **New Features**
  * Added email-delivery tracking (`email_sent_at`) for attendee invoices, credit notes, platform-fee invoices, and referral payout statements.
  * Enabled automatic recovery by regenerating missing PDFs before sending payout statements, invoices, and credit notes.
  * Introduced periodic backstop to re-send undelivered invoice-related emails and broadened referral payout redispatch to cover missing/undelivered statements.
* **Bug Fixes**
  * Successful sends now persist delivery timestamps (idempotently), reducing duplicate or missed email deliveries.
  * Delivery sweeps now safely target only undelivered items and keep unrelated processing moving.


biagiodistefano:
### Code review

Found 1 issue:

1. The payout-statement delivery path is the only one of the four that did **not** get a PDF self-heal, so it can silently lose the document the PR set out to protect. The widened sweep now selects a statement with `email_sent_at IS NULL`, which includes a statement whose row committed but whose PDF was never written (worker crash between the `transaction.atomic()` block and the out-of-transaction `pdf_file.save()` in `generate_payout_statement` — same window the other paths now self-heal). `generate_payout_statement` returns the existing PDF-less statement without regenerating it, then `_send_payout_statement_email` calls `send_email(attachment_storage_path=statement.pdf_file.name)` with an empty name. `send_email` skips a falsy `attachment_storage_path`, so the referrer gets the statement email with **no PDF attached**, and `statement.mark_email_sent()` then marks it delivered permanently. The attendee-invoice, credit-note, and platform-fee paths all added `ensure_*_pdf_exists()` before sending; the payout path needs an equivalent `ensure`/regenerate step before `_send_payout_statement_email` accesses `pdf_file.name`. (Edge case — requires the crash window — but it defeats the PR's purpose for this path.)

https://github.com/letsrevel/revel-backend/blob/06519b5e418333d0a265f6c23a26af34677cabb5/src/accounts/tasks/payouts.py#L279-L286

PR #650, opened 2026-07-09
chore(ci): cache apt deps, time-balance test shards, notify Discord on publish

biagiodistefano:
## Why

Analysis of the last 4 PR runs showed the test suite itself is fine (~3–4 min pytest per shard); the slow runs (11–15 min wall-clock vs a ~6 min healthy baseline) were caused by CI overhead, not test distribution:

- an `apt install` mirror stall cost one shard **5m34s** (vs the usual ~35s)
- `.test_durations` was never committed, so pytest-split fell back to count-based splitting (~46s straggler spread, 3m05s–3m51s)
- ~1.5–2 min of duplicated setup per job

## What

- **Cache apt packages** (`awalsh128/cache-apt-pkgs-action`) in both the `checks` and `test` jobs, trimmed to `gdal-bin libgdal-dev gettext` — the dropped packages (`curl gnupg jq unzip software-properties-common`) ship preinstalled on `ubuntu-24.04` runners. Removes ~30–60s baseline per job and the mirror-flake tail.
- **Rebuild ldconfig after cache restore.** GeoDjango resolves gdal/geos via `ctypes.util.find_library` → the ldconfig cache; a cache-hit restore places the `.so` files without dpkg's ldconfig trigger. (The `GDAL_LIBRARY_PATH` env export is vestigial — Django reads it from settings, which don't define it.)
- **Time-balanced sharding**: commit `.test_durations` (5,811 entries) so pytest-split splits by runtime, and add `make store-durations` to refresh it. Staleness degrades gracefully — tests missing from the file are distributed by count.
- **Discord notification** when a release image is published (`publish.yaml`).

## Notes

- First run on this branch installs apt deps for real and seeds the cache; hits kick in from the second run onward.
- Expected steady state: ~5–5.5 min end-to-end, shard spread down from ~46s to a few seconds.
- Durations refresh is deliberately manual (`make store-durations` + commit) — automating it in the nightly audit would drag docker + a full suite run into a deps-only workflow for marginal gain.

## Summary by CodeRabbit

* **New Features**
  * Release builds now publish an additional version-specific Docker image tag in GHCR alongside `latest`.
  * Publishing a release can optionally notify Discord with the image version and release link (safely skipped when not configured).

* **Chores**
  * CI now uses a shared backend setup routine across checks/tests, with optional translation compilation.
  * Added a `store-durations` Make target to refresh persisted pytest-split timing data.
  * Added a scheduled/manual workflow to regenerate `.test_durations` and open or update a PR when needed.


biagiodistefano:
### Code review

No blocking issues found. Ran an independent pass (bugs, CLAUDE.md compliance, git history, prior-PR comments, code-comment invariants) — nothing new. The `GDAL_LIBRARY_PATH` step dropped in the composite-action refactor is safe: nothing reads that env var (GeoDjango resolves the library via `ldconfig`), so its removal is inert.

Accounting for CodeRabbit's three findings — all are legitimate hardening, none blocking:

1. **curl injection hygiene in `publish.yaml`** — low practical risk here: `version` derives from the release tag but is gated by the earlier tag-vs-`pyproject.toml` consistency check, and `github.event.release.html_url` is GitHub-generated. Still worth moving both expressions into `env:` as defense-in-depth.

https://github.com/letsrevel/revel-backend/blob/ffd4a60dedc9c3513be5cf34d7b9cc14d4c92372/.github/workflows/publish.yaml#L84-L91

2. **Missing `concurrency:` block in `refresh-test-durations.yaml`** — real but low-probability (a manual dispatch overlapping the weekly cron), and the script is idempotent (skip-if-unchanged, force-push, open-PR-only-if-none). A `concurrency` group with `cancel-in-progress: true` would still be cleaner.

https://github.com/letsrevel/revel-backend/blob/ffd4a60dedc9c3513be5cf34d7b9cc14d4c92372/.github/workflows/refresh-test-durations.yaml#L16-L25

3. **Broad App-token scope in `refresh-test-durations.yaml`** — `permission-contents: write` / `permission-pull-requests: write` inputs would scope the minted token to least privilege.

https://github.com/letsrevel/revel-backend/blob/ffd4a60dedc9c3513be5cf34d7b9cc14d4c92372/.github/workflows/refresh-test-durations.yaml#L32-L37

None of the three block merge; treat them as optional follow-ups.

PR #667, opened 2026-07-10
fix: de-duplicate search results across M2M joins (#664)

### LINKED ISSUE #664: Event search returns each event once per matching JOIN row (missing .distinct())
Found by the E2E suite (frontend #589 branch): the events list endpoint returns duplicate rows whenever `search` is used — one copy per joined row (tag count matches the multiplier):

```
GET /api/events/?search=Sunset    → Summer Sunset Music Festival ×3 (3 tags)
GET /api/events/?search=Classical → 6 rows, 2 unique
GET /api/events/?search=Wellness  → 3 rows, 1 unique
```

Frontend impact: the events list renders each result keyed by id, so duplicated ids break the list rendering entirely (blank list, "3 events found" with no cards) — the guest search journey is broken while this stands.

Likely a missing `.distinct()` on the search queryset where it joins tags (and possibly other M2M fields).

The FE E2E spec `j01-guest/browse-events.spec.ts` has the search test marked `test.fixme` referencing this issue — flip it back on when fixed.

biagiodistefano:
Fixes #664.

## Problem
`ninja_extra`'s `Searching` applies the search filter as a single `queryset.filter(Q(...) | ...)` with **no `.distinct()`**. When a `search_field` spans a *multiplying* relation (M2M / reverse FK, e.g. `tags__tag__name`), the join emits one row per related match, so each object was returned **once per matching row**.

For the events list this meant an event with 3 tags appeared 3× when `search` was used (`Summer Sunset Music Festival ×3`), which breaks id-keyed rendering on the frontend (blank list / "3 events found" with no cards) — the guest search journey was broken.

## Fix
Add a small shared `DistinctSearching(Searching)` (in `common/controllers/`) that appends `.distinct()` after the base filter, and use it on **every** search endpoint whose `search_fields` traverse an M2M / reverse relation:

- events discovery (`list_events`) — the reported endpoint
- event series (`list_event_series`)
- organizations (`list_organizations`)
- dashboard: organizations, events, series
- org questionnaires (`list_org_questionnaires` — `events__name`, `event_series__name` are both M2M)

Forward-FK search endpoints (`event__name`, `organization__name`, …) keep the base `Searching` — a many-to-one join can't multiply.

## Test
Regression test `test_list_events_search_no_duplicates_with_tags` creates a 3-tag event and asserts it appears exactly once when searched. Verified it fails on the pre-fix code and passes with the fix.

## Follow-up (frontend, separate repo)
The FE E2E spec `j01-guest/browse-events.spec.ts` has the search test marked `test.fixme` referencing #664 — flip it back on in a frontend PR once this ships.

### DIFF
--- a/src/common/controllers/__init__.py
+++ b/src/common/controllers/__init__.py
@@ -1,5 +1,6 @@
 from .base import UserAwareController
 from .media import MediaValidationController
+from .searching import DistinctSearching
 from .tags import TagController
 
-__all__ = ["MediaValidationController", "TagController", "UserAwareController"]
+__all__ = ["DistinctSearching", "MediaValidationController", "TagController", "UserAwareController"]
--- a/src/common/controllers/searching.py
+++ b/src/common/controllers/searching.py
@@ -1,0 +1,27 @@
+import typing as t
+
+from django.db.models import QuerySet
+from ninja_extra.searching import Searching
+
+
+class DistinctSearching(Searching):
+    """`Searching` that de-duplicates results after filtering.
+
+    The default `Searching` builds an OR across `search_fields` and applies it as a single
+    `queryset.filter(...)`. When any search field spans a multiplying relation (a many-to-many
+    or reverse foreign key, e.g. `tags__tag__name`), the join emits one row per related match,
+    so an object appears once per matching related row. Appending `.distinct()` collapses those
+    duplicates back to one row per object.
+
+    Use this instead of `Searching` whenever `search_fields` traverses an M2M/reverse relation;
+    forward FK joins cannot multiply and do not need it.
+    """
+
+    def searching_queryset(
+        self, items: t.Union[QuerySet[t.Any], t.List[t.Any]], searching_input: "Searching.Input"
+    ) -> t.Union[QuerySet[t.Any], t.List[t.Any]]:
+        """Apply the base search filter, then de-duplicate querysets with `.distinct()`."""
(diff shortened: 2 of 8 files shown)


biagiodistefano:
### Code review

Found 1 issue:

1. `DistinctSearching.searching_queryset` applies `.distinct()` unconditionally, even when no `?search=` term is present. `ninja_extra`'s base `Searching.searching_queryset` no-ops (`return items` unchanged) when there are no search terms, but this override always calls `.distinct()` on the returned queryset. Because `@searching` runs before `@paginate`'s COUNT, this defeats the "materialize IDs to avoid expensive COUNT(*) on complex DISTINCT subquery" optimization in `list_events` — and the equivalent pattern behind `dashboard_organizations`/`dashboard_events`/`dashboard_event_series` — for *every* request to those endpoints, not just searches. The returned queryset there is already join-free and unique (`filter(id__in=event_ids)`), so distinct is only needed when the search filter actually re-adds the tags join. Consider gating on a real search term (mirror the base class's `if self.search_fields and search_terms:`, or short-circuit when `result is items`). (`list_organizations`/`list_event_series`/`list_org_questionnaires` already call `.distinct()` in their bodies, so for those it's merely redundant.)

https://github.com/letsrevel/revel-backend/blob/0c5f511ae8d4fa3af7cf181cd014b178c53bb974/src/common/controllers/searching.py#L20-L27

PR #676, opened 2026-07-11
feat: expose org name/slug/logo on public org-token preview (#675)

### LINKED ISSUE #675: Enrich public org-token details with organization name/slug/logo (FE /join/org page is blind)
Backend half of frontend letsrevel/revel-frontend#601.

## Problem

`GET /api/organizations/tokens/{token_id}` (the public, pre-claim token preview) returns `OrganizationTokenSchema` where `organization` is a bare UUID. The frontend invitation page (`/join/org/[token_id]`) needs to show **which org the user is being invited to** (name, ideally slug + logo) before they claim — and there is no public organization-by-id endpoint to resolve the UUID, so the FE literally cannot render the page correctly today (it currently SSR-500s; see FE #601).

## Proposal

Extend the token-details response with a nested minimal org object (or flat fields), e.g.:

```json
{ "organization": { "id": "...", "name": "...", "slug": "...", "logo": "... | null" }, ... }
```

or additive flat fields (`organization_name`, `organization_slug`, `organization_logo`) if changing the `organization` field type is too breaking — the same schema backs the org-admin token list, so additive is probably safer.

Care: this endpoint is unauthenticated — expose only what a public invitee may see (name/slug/logo of the inviting org seems fine given they hold a valid token; consider what to return for expired/used tokens).

## Impact

- Unblocks fixing FE #601 (`/join/org/[token_id]` SSR 500 for every token).
- Unblocks the E2E v2 j01 `token-preview` spec (frontend #590, Phase 2 group g).

biagiodistefano:
Backend half of frontend letsrevel/revel-frontend#601. Fixes #675.

## Problem

`GET /api/organizations/tokens/{token_id}` (the public, pre-claim token preview) returned `OrganizationTokenSchema` where `organization` is a bare UUID. The `/join/org/[token_id]` page needs to show which org is inviting (name/slug/logo) before the user claims, and there is no public org-by-id endpoint to resolve the UUID — so the page SSR-500s today.

## Change

Additive, public-safe fields on the token schema, resolved from the token's organization:

- `organization_name: str`
- `organization_slug: str`
- `organization_logo_url: str | null` (logo thumbnail, public/unsigned)

This matches the existing flat `organization_*` convention used by the membership/subscription schemas (`subscription.py`), rather than a nested object. `organization` stays a bare UUID, so the **org-admin token list/create** endpoints (which share `OrganizationTokenSchema`) keep their existing contract — non-breaking.

Only name/slug/logo are exposed (no stripe/fee/contact fields) since the endpoint is unauthenticated. Expired/used tokens already 404 at the service layer (`get_organization_token` filters them out), so no org data leaks for invalid tokens.

Also `select_related("organization")` on the admin token list queryset to avoid an N+1 introduced by the new resolvers (the public endpoint already select_related's via the service).

## Testing
- `uv run pytest src/events/tests/test_controllers/test_token_endpoints.py src/events/tests/test_controllers/test_organization_admin/test_tokens.py`

The FE client should be regenerated (`pnpm generate:api`) against the updated OpenAPI spec after this merges.

https://claude.ai/code/session_01Bn3RqswtYv5TjXiqn5J51N

### DIFF
--- a/src/events/controllers/organization_admin/tokens.py
+++ b/src/events/controllers/organization_admin/tokens.py
@@ -101,7 +101,9 @@ class OrganizationAdminTokensController(OrganizationAdminBaseController):
         - Verify staff invitation tokens are properly restricted
         """
         organization = self.get_one(slug)
-        return params.filter(models.OrganizationToken.objects.filter(organization=organization)).distinct()
+        return params.filter(
+            models.OrganizationToken.objects.filter(organization=organization).select_related("organization")
+        ).distinct()
 
     @route.post(
         "/tokens",
--- a/src/events/schema/organization.py
+++ b/src/events/schema/organization.py
@@ -26,6 +26,7 @@ from .mixins import (
     SocialMediaSchemaEditMixin,
     SocialMediaSchemaRetrieveMixin,
     TaggableSchemaMixin,
+    get_image_field_url,
 )
 
 
@@ -300,6 +301,13 @@ class StaffAddSchema(Schema):
 
 
 class OrganizationTokenSchema(ModelSchema):
+    # Additive, public-safe org details so the pre-claim token preview (unauthenticated)
+    # can render which organization the invitee is joining without a second lookup.
+    # ``organization`` stays a bare UUID to keep the org-admin token list contract unchanged.
+    organization_name: str
+    organization_slug: str
+    organization_logo_url: str | None = None
+
     class Meta:
         model = models.OrganizationToken
         fields = [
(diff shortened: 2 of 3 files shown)


florimond-collette:
A quick review of the merged changes:

1. **Performance & N+1:** The addition of .select_related('organization') on the list query is correct and prevents N+1 queries for the new resolvers. Since issuer and membership_tier are serialized as bare UUIDs and do not use nested schemas, they use the local database columns (issuer_id, membership_tier_id) and do not trigger extra queries.
2. **Security & Privacy:** Restricting issuer and membership_tier to bare UUIDs is a great design choice to keep the unauthenticated preview endpoint safe from user/metadata harvesting.

I will open a tracking issue to address a minor UX improvement: exposing membership_tier_name so the pre-claim /join/org landing page can render which specific tier a user is being invited to join.

PR #685, opened 2026-07-12
fix: guard org-admin routes on real PermissionMap keys (#683)

### LINKED ISSUE #683: OrganizationPermission guards reference keys absent from PermissionMap (manage_organization / manage_events) — deny all non-owner staff
Surfaced while building the frontend @p2 E2E suite (letsrevel/revel-frontend#591, group d — `j09/permission-gating`).

Several `OrganizationPermission("<key>")` route guards reference permission keys that are **not defined in `PermissionMap`** (`src/events/models/organization.py:462`):

- `resources.py` DELETE → `manage_organization`
- `revenue.py` (financials) → `manage_organization`
- `discount_codes.py` → `manage_events`

Neither `manage_organization` nor `manage_events` exists in `PermissionMap` (the real keys are `manage_organization`? no — the map has `manage_event`, `manage_tickets`, `manage_members`, `edit_organization`, etc., but not these two).

**Effect:** `OrganizationStaff.has_permission("manage_organization"|"manage_events", ...)` can never resolve to a granted value, so those routes **deny every non-owner staff member regardless of the permissions they were granted** — effectively making resource-deletion, financials/revenue, and discount-code management owner-only, even when a staff grant was intended to allow them.

**Ask:** either
1. add `manage_organization` / `manage_events` to `PermissionMap` (with sensible defaults) and surface them in the staff-permission editor, or
2. change these guards to existing map keys (e.g. `manage_event`, `edit_organization`).

If owner-only is actually intended for these routes, use `IsOrganizationOwner` explicitly instead of a permission key that silently never matches — the current form reads like a bug.

biagiodistefano:
Fixes #683.

## Problem

Three `OrganizationPermission("<key>")` route guards referenced permission keys **not present in `PermissionMap`** (`events/models/organization.py`). `PermissionMap` is the canonical, enumerated set of staff-permission keys — both the Django admin staff form (`admin/base.py`) and the frontend editor (generated from this schema) render checkboxes *only* for keys in the map. Because no legitimate path can grant a key that isn't in the map, `OrganizationStaff.has_permission(...)` always returned `False`, so these routes **silently denied every non-owner staff member regardless of the permissions they were granted** — effectively making them owner-only, but via a guard that reads like a bug.

| Route | Before | After |
|---|---|---|
| `resources.py` DELETE resource | `manage_organization` ❌ | `edit_organization` (matches create/update in the same controller) |
| `discount_codes.py` (whole controller) | `manage_events` ❌ | `manage_tickets` (discount codes are ticket pricing) |
| `revenue.py` financials/VAT reports (whole controller) | `manage_organization` ❌ | `IsOrganizationOwner()` (owner-only, made explicit) |

### Why these choices

- **Resources DELETE → `edit_organization`**: create/update in the same controller already use `edit_organization`; deleting is the same tier of operation.
- **Discount codes → `manage_tickets`**: semantically closest existing key. Note this widens access: `manage_tickets` defaults to `True`, so staff who were (incorrectly) locked out can now manage codes as intended.
- **Financials → owner-only**: financial/VAT reports are owner territory; `IsOrganizationOwner()` states that explicitly instead of relying on a key that silently never matches.

## Prevention: compile-time guard against bogus permission keys

To stop this class of bug from recurring, the permission-guard constructors are now **type-constrained to real `PermissionMap` keys**, so mypy (and any mypy/pyright-backed IDE) rejects a non-existent key at the call site.

- Added `PermissionKey = t.Literal[...]` next to `PermissionMap` (exported from `events.models`).
- Added a `PermissionMapPermission` base whose `__init__(action: PermissionKey)` narrows the type; reparented the map-resolving permission classes onto it: `OrganizationPermission`, `EventPermission`, `EventSeriesPermission`, `QuestionnairePermission` (events) and `PollPermission` (polls). The fixed/special-action classes (`IsOrganizationOwner`, `PotluckItemPermission`, `CanPurchaseTicket`, …) — whose actions are deliberately *not* map keys (`is_owner`, `create_potluck_item`, …) — stay on `RootPermission` untouched.
- Added a drift-guard test asserting `PermissionKey` lists exactly the `PermissionMap` fields, since mypy can't verify the literal against the model itself.

This is a type-only change with no runtime behavior difference. Verified that `OrganizationPermission("manage_tickets_typo")` now produces a mypy error listing every valid key.

## Tests

- Discount-code suite now grants/denies the real `manage_tickets` key (added a `no_tickets_staff` fixture for the deny case, since `manage_tickets` defaults to `True`).
- Financials: renamed the member-deny test and added a **staff-deny** test proving owner-only.
- Resources: added a regression test that a staff member granted `edit_organization` **can** delete a resource — the exact scenario the bug broke.
- New `test_permission_key_sync.py` guards `PermissionKey` ≡ `PermissionMap`.

The pre-existing tests passed only because they injected the bogus key straight into the JSON blob (`perms["default"]["manage_events"] = True`), which no real UI/API does — so they never caught the bug.

`make format lint mypy` all pass.

https://claude.ai/code/session_01Bn3RqswtYv5TjXiqn5J51N

## Summary by CodeRabbit

* **Access Control**
  * Discount-code management now requires the **Manage tickets** permission.
  * Organization staff with **Edit organization** permission can delete additional resources.
  * Financial information is restricted to **organization owners**; members and staff are denied.

* **Bug Fixes**
  * Improved and standardized authorization behavior across organization-admin endpoints, with updated automated coverage to verify the correct permission requirements and forbidden access responses.

### DIFF
--- a/src/events/controllers/organization_admin/discount_codes.py
+++ b/src/events/controllers/organization_admin/discount_codes.py
@@ -22,7 +22,7 @@ from .base import OrganizationAdminBaseController
     auth=I18nJWTAuth(),
     tags=["Organization Admin"],
     throttle=WriteThrottle(),
-    permissions=[OrganizationPermission("manage_events")],
+    permissions=[OrganizationPermission("manage_tickets")],
 )
 class OrganizationAdminDiscountCodesController(OrganizationAdminBaseController):
     """Organization discount code management endpoints."""
--- a/src/events/controllers/organization_admin/resources.py
+++ b/src/events/controllers/organization_admin/resources.py
@@ -95,7 +95,7 @@ class OrganizationAdminResourcesController(OrganizationAdminBaseController):
         "/resources/{resource_id}",
         url_name="delete_organization_resource",
         response={204: None},
-        permissions=[OrganizationPermission("manage_organization")],
+        permissions=[OrganizationPermission("edit_organization")],
     )
     def delete_resource(self, slug: str, resource_id: UUID) -> tuple[int, None]:
         """Delete a resource from the organization."""
--- a/src/events/controllers/organization_admin/revenue.py
+++ b/src/events/controllers/organization_admin/revenue.py
@@ -12,7 +12,7 @@ from common.authentication import I18nJWTAuth
 from common.models import FileExport
 from common.throttling import ExportThrottle, UserDefaultThrottle
 from events import schema
-from events.controllers.permissions import OrganizationPermission
+from events.controllers.permissions import IsOrganizationOwner
 from events.exceptions import InvalidPeriodError
 from events.schema.export import FileExportSchema
 from events.schema.financials import OrganizationFinancialsSchema
@@ -25,7 +25,7 @@ from .base import OrganizationAdminBaseController
     "/organization-admin/{slug}",
     auth=I18nJWTAuth(),
     tags=["Organization Admin"],
-    permissions=[OrganizationPermission("manage_organization")],
+    permissions=[IsOrganizationOwner()],
 )
 class OrganizationAdminRevenueController(OrganizationAdminBaseController):
     """Generate and poll downloadable revenue & VAT report bundles."""
--- a/src/events/controllers/permissions.py
+++ b/src/events/controllers/permissions.py
@@ -30,7 +30,20 @@ class RootPermission(BasePermission):
         return True
 
 
-class EventSeriesPermission(RootPermission):
+class PermissionMapPermission(RootPermission):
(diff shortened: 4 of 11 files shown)


biagiodistefano:
### Code review

florimond-collette:
LGTM! N+1 patterns are well-handled with the new `select_related`/`prefetch_related` on the enriched preview endpoints, and all permission guards align correctly with `PermissionMap`.

Unrelated: `uvloop` currently blocks developer setup on Windows — worth making it conditional in a future chore:
```toml
"uvloop>=0.22.1; sys_platform != 'win32'"
```

PR #860, opened 2026-08-07
fix(notifications): don't duplicate the Payment Instructions heading

### LINKED ISSUE #859: emails: pending-payment email renders "Payment Instructions" heading twice
Found during the revel-frontend#812 pre-merge smoke pass (offline-reservation flow, "Ticket Pending - Classical Music Evening" email).

**Symptom:** the HTML email template emits its own `<h3>Payment Instructions</h3>` and then renders the organizer-supplied markdown, which (at least in the seeded data, and plausibly in real orgs copying the pattern) itself starts with `## Payment Instructions` — so the heading appears twice in a row. The markdown itself renders correctly (`<h2>`, `<strong>`, `<ul>` — not raw markup); it's purely the duplicated heading.

**Fix sketch:** either drop the template's hardcoded heading and let the organizer content own its structure, or strip/demote a leading heading from the organizer markdown when the template already provides one.

https://claude.ai/code/session_01JDXmzqVQzDQTBwT98KL9Wf

biagiodistefano:
Fixes #859.

## Problem

The pending-ticket notification renders its own **Payment Instructions** heading and then the organizer-supplied markdown, which commonly starts with its own `## Payment Instructions` (that's what the bootstrap seed does, and it's a natural pattern for organizers to copy) — so the heading appeared twice in a row.

## Fix

Neither of the issue's two sketches works alone:

- Dropping the template heading breaks orgs that write bare text with no heading of their own (e.g. the showcase seed's `"Please transfer to IBAN: XX1234567890"`) — the card would be unlabelled.
- Demoting the organizer's heading still leaves two "Payment Instructions" lines.

So: keep the template heading (it's the card label, consistent with the Event Details / Ticket Information cards) and strip a **leading** heading from the organizer content. The organizer's first heading is redundant by construction — it sits directly under our label — while any later headings, which are real structure, are preserved.

New `strip_leading_heading` filter in `common/templatetags/markdown_tags.py`, applied on all four `ticket_created` channels (email HTML, email text, in-app, Telegram) — the duplication was identical in each, and the issue's HTML repro is just where it was spotted.

Only ATX headings (`## Title`) are handled; setext (`Title\n====`) is left with a `# ponytail:` note.

## Tests

- `src/common/tests/test_markdown_tags.py` — new, covers the filter: every heading level, leading blank lines, only-the-first-heading, heading-only content, and the non-heading cases (`#1 rule`, mid-text `#`, heading below other content).
- `TestPendingTicketPaymentInstructions` in `test_ticket_templates.py` — parametrized over all four channels: leading heading not duplicated, bare instructions preserved under the heading, missing instructions still fall back to the contact-the-organizer copy.

```
27 passed  (new tests)
85 passed  (test_ticket_templates.py + test_ticket_templates_branding.py)
```

`make format` / `lint` / `mypy` / `i18n-check` all clean. No new translatable strings, so no `.po` churn.

https://claude.ai/code/session_014YbEH5FH1ET6EPZCABR2JV

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved payment instructions in ticket-created notifications by removing redundant leading Markdown headings.
  * Applied consistent formatting across email, in-app, and Telegram notifications.
  * Removed only the initial heading while preserving later headings, indentation, and headingless instructions.
  * Preserved fallback contact information when payment details are unavailable.

* **Tests**
  * Added coverage for heading levels, whitespace handling, heading-only content, hashtag-like text, and notification rendering across supported channels.

### DIFF
--- a/src/common/templatetags/markdown_tags.py
+++ b/src/common/templatetags/markdown_tags.py
@@ -11,6 +11,8 @@ from common.fields import render_markdown
 
 register = template.Library()
 
+_LEADING_ATX_HEADING_RE = re.compile(r"\A\s*#{1,6}[ \t]+\S[^\n]*(?:\n|\Z)")
+
 
 @register.filter(is_safe=True)
 def markdown(value: str | None) -> str:
@@ -35,6 +37,27 @@ def markdown(value: str | None) -> str:
     # render_markdown calls nh3.clean() with a strict allowlist, so the output
     # is sanitized and safe to mark as such.
     return mark_safe(render_markdown(value))
+
+
+@register.filter
+def strip_leading_heading(value: str | None) -> str:
+    """Drop a leading markdown heading so it doesn't duplicate a template-provided title.
+
+    Organizer-authored markdown (e.g. manual payment instructions) is rendered inside
+    cards that already carry their own heading. When the author starts their text with
+    ``## Payment Instructions``, the heading shows up twice in a row. This filter removes
+    that first heading; anything else in the content is untouched.
+
+    Usage:
+        {% load markdown_tags %}
+        {{ context.manual_payment_instructions|strip_leading_heading|markdown }}
+    """
+    if not value:
+        return ""
+
+    # ponytail: only ATX headings ("## Title"). Setext ("Title\n====") is rare in
+    # organizer content; extend here if it shows up.
+    return _LEADING_ATX_HEADING_RE.sub("", value).lstrip()
 
 
 @register.filter
--- a/src/common/templatetags/markdown_tags.py
+++ b/src/common/templatetags/markdown_tags.py
@@ -11,7 +11,9 @@ from common.fields import render_markdown
 
 register = template.Library()
 
-_LEADING_ATX_HEADING_RE = re.compile(r"\A\s*#{1,6}[ \t]+\S[^\n]*(?:\n|\Z)")
+# A leading ATX heading plus the blank lines under it. The trailing `(?:[ \t]*\n)*`
+# stops at the first line with content, so that line's own indentation survives.
+_LEADING_ATX_HEADING_RE = re.compile(r"\A\s*#{1,6}[ \t]+\S[^\n]*(?:[ \t]*\n)*")
 
 
 @register.filter(is_safe=True)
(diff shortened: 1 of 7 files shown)


biagiodistefano:
### Code review

Found 1 issue:

1. When the organizer's entire instructions are a single ATX-heading line, `strip_leading_heading` returns `""` — the regex's `(?:\n|\Z)` lets it consume the whole string. All four templates gate the `{% if %}` on the *raw* value, so the heading still renders, the body is empty, and the `{% else %}` fallback never fires. The payment info silently disappears. Verified by rendering `notifications/email/ticket_created.txt` with `manual_payment_instructions="# Please e-transfer to IBAN AT12 3456 7890"`:

```
'Payment Instructions:\n\n\n\nView event details: https://e.com'
fallback present: False
```

https://github.com/letsrevel/revel-backend/blob/821015ff4cbb3ac02e8317696ac6ad514ced5fec/src/common/templatetags/markdown_tags.py#L59-L61

https://github.com/letsrevel/revel-backend/blob/821015ff4cbb3ac02e8317696ac6ad514ced5fec/src/notifications/templates/notifications/email/ticket_created.html#L34-L37

`test_heading_only_content_becomes_empty` pins this as intended filter behavior, but the template-level consequence — a labelled card with nothing in it — isn't covered by any test. Options: leave the value untouched when stripping would empty it, or move the fallback decision onto the filtered value.

biagiodistefano:
Fixed in a5186da5. Both defects came from the same two lines, so one restructure covers them:

- **Content wipeout**: `strip_leading_heading` now returns the original value when stripping would empty it, so a single-line `# ...` instruction survives instead of leaving a labelled card with no body.
- **Unconditional `.lstrip()`** (raised by CodeRabbit, scored just under my reporting bar but real): non-matching content is now returned verbatim, and the regex consumes only the blank lines *under* a heading — `(?:[ \t]*\n)*` stops at the first line with content, so that line keeps its own indentation.

Verified a 4-space block still renders as `<pre><code>` both with and without a leading heading. New tests pin heading-only content, indentation with and without a heading, and a template-level case across all four channels asserting the one-liner reaches the recipient.

Issue #880, opened 2026-08-11
perf: endpoint performance umbrella — DISTINCT planner bombs, openapi-json regen, admin inlines, my_permissions caching, confirm-payment stall


biagiodistefano:
## Context

`/my-tickets` (frontend) is reported as unreasonably slow. A static scan of the two backend endpoints the page calls — `GET /api/dashboard/tickets` and `GET /api/series-passes/me` — found no unconditional N+1, but three real inefficiencies on the tickets endpoint. None of them obviously explains multi-second latency for a typical user, so a Silk profiling session is still needed to confirm which (if any) dominates; this issue tracks the statically-confirmed findings.

Baseline is otherwise solid: `Ticket.objects.full()` covers everything `UserTicketSchema` touches, `pdf_url`/`pkpass_url` are in-process HMAC signing, `apple/google_pass_available` are settings checks, and `/series-passes/me` serializes entirely from prefetched rows.

### 1. Conditional N+1: `staff_members` M2M query per ticket

`TicketTierSchema.resolve_total_available` → `Event.can_user_see_capacity(viewer)` → `bypasses_visibility_settings()` → `self.organization.staff_members.all()` (`src/events/models/event.py:558`).

`show_capacity` defaults to `True` (short-circuits before the M2M), but for every ticket whose event **hides capacity**, serialization issues one `staff_members` query per row — the per-instance `_visibility_bypass_cache` is useless here because each ticket row carries its own `event` instance. A 20-ticket page = up to 20 extra queries.

Fix direction: prefetch `tier__event__organization__staff_members` in `Ticket.full()`, or resolve the bypass once per request instead of per event instance.

### 2. Conditional N+1: `resolve_seat_pricing` per ticket row

`TicketTierSchema.resolve_seat_pricing` (`src/events/schema/ticket_tier.py:195`) costs one `PriceCategory` query per category-priced tier. The docstring's justification ("an event has a handful of tiers at most") holds on the tier-list endpoint, but in my-tickets the schema serializes **per ticket row**, so a page of category-priced seated tickets costs up to 20 queries.

### 3. `.distinct()` is a semantic no-op that costs real work

`dashboard_tickets` (`src/events/controllers/dashboard.py:213`) ends with `.distinct()`, but every filter and search field on the endpoint (`status`, `tier__payment_method`, `event__end`, search over `event__name`/`event__description`/`tier__name`) is a single-valued forward relation — no join can fan out rows.

[...]

PR #882, opened 2026-08-11
perf(api): remove DISTINCT planner bombs, cache openapi & my-permissions, fix admin inline N+1s (#880)

biagiodistefano:
Implements all findings from #880 (endpoint performance umbrella). Root-cause analysis, measurements, and the invalidation spec live in that issue.

### 1. Remove `.distinct()` from five list endpoints
`dashboard_tickets` / `dashboard_rsvps` / `dashboard_invitation_requests`, admin `list_tickets` / `list_ticket_tiers`.

`EXPLAIN ANALYZE` showed the DISTINCT over the wide `full()` select costs **~600ms–700ms of Postgres planning time per request** (execution: 1.5ms) — ~2.6s on demo hardware — while deduplicating nothing: every filter, search field, annotation, and ordering on these endpoints is a to-one join (verified exhaustively per field in #880). Git history confirms the `.distinct()` was never load-bearing here — the endpoints were born with it; the real duplicate-rows problem lived on the M2M-joining dashboard org/event endpoints and was fixed properly by the UNION-of-IDs refactor in #63.

**Regression tests** create the would-be fan-out conditions (tiers restricted to two membership tiers, search terms matching multiple joined fields simultaneously, multi-value status filters, `source` + annotation ordering) and assert exact counts, exact ID sets, no duplicate IDs, and no `SELECT DISTINCT` in the captured SQL. Measured locally: 615ms → **15ms**.

### 2. Memoize OpenAPI schema generation
django-ninja rebuilds the full schema (544 component schemas, 854KB) on every `/api/openapi.json` request — a public, unthrottled view (the API-level throttles don't apply to ninja's plain Django docs views), so each hit burned ~2.4s of CPU while pinning a gunicorn worker + PgBouncer connection. `CachedSchemaNinjaExtraAPI` builds it once per process; the schema depends only on registered routes and the mount prefix, both fixed after startup. The collision-guard tests reset the memo so they still observe real generation; `dump_openapi` is unaffected (single call, its monkeypatch active during the only build).

### 3. Fix admin Event-change-page inline widget N+1s (p95 4.2s)
The inlines rendered plain `<select>`s over entire tables — `seat` over every `VenueSeat` in the DB with one `__str__` query per option per row (+ the always-rendered empty form), `tiers` over every TicketTier, `user` over every registered user. Now `raw_id_fields` / `autocomplete_fields` on `TicketInline`, `TicketTierInline`, `EventInvitationInline`, `WaitlistOfferInline`.

### 4. Cache `my_permissions` (~147k req/week)
New `events/service/permission_snapshot.py`: per-user 60s-TTL cache of the rendered payload, versioned key (`my_permissions:v1:{user_id}`), plain `model_dump(mode="json")` dict (ninja re-validates on the way out, so a corrupt entry 500s loudly instead of serving plausible-but-wrong data), **fail-open on every cache op**.

Safety model (verified in #880): no server-side authorization ever reads this payload — every admin/staff/purchase endpoint re-checks the DB per request — so staleness is UI-only and bounded by the TTL. Explicit invalidation exists at exactly the two interactive grant sites where a stale miss would 403 the page the FE navigates to next (`create_organization`, `claim_invitation`), via `transaction.on_commit` — never in-transaction, which would reopen the invalidate-then-repopulate race under `ATOMIC_REQUESTS`. Tests pin the deferred-to-commit ordering, rollback safety, fail-open, and cache-hit/fresh-build equivalence.

Includes the free trim: `values_list` instead of hydrating full Organization rows for the staff map.

### 5. Bound Redis cache ops with socket timeouts
The built-in `RedisCache` client had no timeouts, so a hung (vs down) Redis would block the global throttle check on every request indefinitely. New `REDIS_SOCKET_CONNECT_TIMEOUT` / `REDIS_SOCKET_TIMEOUT` (default 1.0s), documented in `.env.example` and the troubleshooting env-var table.

## Not included (still open in #880)
- The two conditional per-row N+1s (`resolve_seat_pricing`, `resolve_total_available` → `staff_members`) — secondary, tens of ms.
- DB `lock_timeout` / `log_lock_waits` hardening for the 48s `confirm_ticket_payment` class — needs a value decision, separate change.

## Verification
- `mypy --strict` clean (1188 files); ruff format + lint clean; i18n / file-length / task-names gates pass.
- ~290 targeted tests pass: new regression suites, permissions, dashboard, admin tickets, org controller, error-response contracts, api tests.
- No migrations.

Closes nothing on its own — #880 stays open for the remaining items above.

https://claude.ai/code/session_017x1LCUdzwpXo2upKb1R1zW

## Summary by CodeRabbit

* **Performance**
  * Improved permission loading with short-term caching and automatic updates after organization or invitation changes.
  * Prevented Redis operations from hanging indefinitely with configurable connection and socket timeouts.
  * Improved dashboard, ticket-list, and API schema performance while preserving unique results.

* **Admin Improvements**
  * Added faster ID selection and user autocomplete for large ticketing and waitlist forms.

* **Documentation**
  * Documented Redis timeout settings, defaults, and configuration examples.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,6 +12,14 @@ DB_PASSWORD=db-password
 DB_NAME=revel
 DB_PORT=5432
 
+# REDIS (cache + Celery broker)
+# REDIS_HOST=localhost
+# REDIS_PORT=6379
+# Per-operation socket timeouts (seconds) for the Django cache client. Bound every
+# cache op so a hung Redis errors out instead of blocking requests indefinitely (#880).
+# REDIS_SOCKET_CONNECT_TIMEOUT=1.0
+# REDIS_SOCKET_TIMEOUT=1.0
+
 APPLE_TEAM_IDENTIFIER=FOOBAR
 
 # Apple Wallet Pass Configuration (optional)
--- a/docs/getting-started/troubleshooting.md
+++ b/docs/getting-started/troubleshooting.md
@@ -197,6 +197,7 @@ Most external services have safe defaults for local development. Here's what act
 |---------|---------|---------|-------|
 | PostgreSQL | `DB_USER`, `DB_PASSWORD`, `DB_NAME`, `DB_PORT` | `revel`, `db-password`, `revel`, `5432` | Provided by Docker Compose (values from `.env.example`) |
 | Redis | `REDIS_HOST` | `localhost` | Provided by Docker Compose |
+| Redis timeouts | `REDIS_SOCKET_CONNECT_TIMEOUT`, `REDIS_SOCKET_TIMEOUT` | `1.0`, `1.0` | Seconds; bound every cache op so a hung Redis errors instead of blocking requests (#880) |
 
 ### Required for specific features
 
--- a/src/api/api.py
+++ b/src/api/api.py
@@ -1,7 +1,10 @@
+import typing as t
+
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.http import Http404, HttpRequest
 from django.utils.translation import gettext_lazy as _
+from ninja.openapi.schema import OpenAPISchema
 from ninja_extra import NinjaExtraAPI
 
 from accounts.controllers.account import AccountController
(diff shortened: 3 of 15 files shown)


biagiodistefano:
### Code review

Five parallel review passes (CLAUDE.md audit, bug scan, git-history check, prior-PR feedback, code-comment invariants) surfaced one sub-threshold consistency nit — `_declared_error_schemas()` wrapped `get_openapi_schema()` in the collision guard without dropping the new process-level memo, so the guard observed a cache hit rather than real generation — already fixed in e9af723d. Notably, git history confirms none of the five removed `.distinct()` calls was ever added as a standalone duplicate-row fix (the real historical dupe fix, #664's `DistinctSearching`, targeted `tags__tag__name` M2M search on different endpoints, untouched here).

PR #916, opened 2026-09-02
fix(security): remediate 12 findings from the vuln-scan sweep

biagiodistefano:
## Summary

Remediates the 12 confirmed findings (4 HIGH, 8 MEDIUM) from a full-codebase security sweep. Every fix is surgical and mirrors patterns already used elsewhere in the codebase. The exploit tests that reproduced each finding are intentionally **not** included in this PR (kept local, per convention); they now fail against this branch, which is the intended signal that the vulnerabilities are closed.

## HIGH

- **Ban evasion (accounts).** Globally-banned emails were accepted on the guest-user creation and password-reset paths, and a banned domain could be converted into a full active account. Now `is_email_globally_banned` gates guest creation and the reset request/redeem pair, and `handle_guest_rsvp` is atomic so a rejected guest leaves no row.
- **Stored XSS in the Django admin (questionnaires).** Admin display methods interpolated attacker-controlled questionnaire/event names via `mark_safe` f-strings. Rewritten with `format_html` / escaped JSON.
- **Malware re-upload bypass (common).** Re-uploading identical flagged bytes rolled back the quarantine and left the file live. The audit query is now scoped by instance/status and `bulk_create` uses `ignore_conflicts`.
- **Series-pass admission bypass (events).** The coverage gate checked `visibility` instead of `event_type`, letting a public pass grant tickets to invitation-only / members-only events. Now gates on `event_type`.

## MEDIUM

- **Stale-membership resource access (events-models).** `AdditionalResource.for_user` now filters membership through `for_visibility()`, so cancelled/banned members lose MEMBERS_ONLY access.
- **Quarantine no-op on non-nullable file fields (common).** For non-nullable file fields (questionnaire uploads) the row/file is now deleted rather than nulled, so quarantine no longer fails silently.
- **Per-user cap TOCTOU (events-service).** Per-user ticket caps are now enforced under the tier/event row lock, closing a concurrent-checkout bypass of `max_tickets_per_user`.
- **Spreadsheet formula injection (events-service).** Attendee and questionnaire exports route user-derived cells through a `sanitize_cell` helper (CWE-1236).
- **MC scoring inflation (questionnaires).** Each multiple-choice question's weight is awarded at most once and the score is clamped to 100.
- **Event status privilege bypass (events-schema).** `status` moved out of `EventEditSchema` into `EventCreateSchema`, so status transitions go only through the `manage_event`-gated endpoint.
- **Image-upload memory DoS (common).** Added a pixel-dimension bound plus a Pillow backstop and removed the triple-buffer EXIF strip.
- **Recurrence generation cap bypass (events-service).** The client `until` override is clamped to the 52-week generation ceiling.

## Also included

Small updates to the `.claude/agents/*` definitions (vuln-analyst, vuln-verifier, tech-debt-assessor, tech-debt-verifier).

## Verification

- `ruff format`, `ruff check`, and `mypy --strict` pass on all 16 changed source files.
- Full suite (local, parallel): **8525 passed**, with only the 12 (uncommitted) exploit tests failing by design.
- One incidental regression from the image fix (`test_prefers_thumbnail_over_logo`, which saves deliberately-fake bytes) was resolved by only enforcing the pixel budget when dimensions are actually readable, preserving the original validator's leniency.

## Note for review

The MC-scoring fix caps each question's contribution rather than requiring an exact-match (all-or-nothing) answer set, because existing evaluator tests assert that selecting one of several correct options still scores full marks. Switching to all-or-nothing would be a grading-semantics change requiring those tests to be updated — flagging in case that stricter behaviour is desired.

## Summary by CodeRabbit

- **Security**
  - Blocked globally banned addresses from password resets, guest RSVPs, and account creation.
  - Added protections against oversized images, spreadsheet formula injection, and unsafe admin HTML.
  - Improved malware quarantine handling and notification reliability.

- **Bug Fixes**
  - Prevented per-user ticket limits from being bypassed during concurrent purchases.
  - Restricted event status changes to the appropriate workflow.
  - Corrected members-only visibility and questionnaire scoring behavior.
  - Improved safety and accuracy across attendee and questionnaire exports.

- **Improvements**
  - Limited recurrence generation horizons and reduced memory usage when processing images.
  - Prevented restricted events from being included in series passes.

### DIFF
--- a/.claude/agents/tech-debt-assessor.md
+++ b/.claude/agents/tech-debt-assessor.md
@@ -1,7 +1,7 @@
 ---
 name: tech-debt-assessor
 description: "Use this agent to assess technical debt and code quality in a bounded region of the Revel codebase. It is the region-scoped **hunter** worker dispatched in parallel by the `/tech-debt` command, and can also be used standalone for an ad-hoc tech-debt review of a specific app, service, or module. For a full-repo assessment, prefer the `/tech-debt` slash command (which fans out many hunters + specialists + a verification pass and scores the whole repo) over invoking this agent directly.\\n\\nExamples:\\n\\n<example>\\nContext: The user wants a full repo health assessment.\\nuser: \"Let's do a tech debt review\"\\nassistant: \"I'll run the /tech-debt command, which fans out region + specialist hunters and produces a scored report.\"\\n<commentary>Full-repo sweep — prefer the /tech-debt command over invoking this agent directly.</commentary>\\n</example>\\n\\n<example>\\nContext: The user just finished a feature and wants a quick debt read on one app.\\nuser: \"Can you assess the tech debt in the questionnaires app I just reworked?\"\\nassistant: \"I'll launch the tech-debt-assessor scoped to questionnaires to produce a focused health report.\"\\n<commentary>Single-area review. Use the Task tool to launch tech-debt-assessor (Mode B).</commentary>\\n</example>"
-model: opus
+model: fable
 color: pink
 memory: project
 ---
--- a/.claude/agents/tech-debt-verifier.md
+++ b/.claude/agents/tech-debt-verifier.md
@@ -1,7 +1,7 @@
 ---
 name: tech-debt-verifier
 description: "Adjudicates a single candidate tech-debt finding produced by the tech-debt-assessor hunter. Independently re-reads the cited code, confirms the debt is real and present-day (and for dead-code candidates, proves no references exist anywhere), and returns a verdict with a 0-100 confidence score. Dispatched in parallel (one per candidate) by the `/tech-debt` command's verification phase. Not intended for standalone use — it expects a single candidate finding as input."
-model: sonnet
+model: opus
 color: pink
 ---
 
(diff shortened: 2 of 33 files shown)


biagiodistefano:
### Code review

Found 1 issue:

1. The `strip_exif` rewrite reintroduces an EXIF/GPS leak for HEIC/HEIF uploads (bug due to `src/common/utils.py` calling `image.save(output, format=_format)` on the original decoded image instead of a fresh `Image.frombytes(...)` copy). Pillow's built-in JPEG/PNG/WEBP encoders only read `exif` from the explicit save kwargs, so those formats still come out clean. But `pillow-heif` (registered via `register_heif_opener()` in `src/revel/settings/base.py`) forwards `img.info` into the encoder: `info = img.info.copy(); info["exif"] = _exif_from_pillow(img); info["xmp"] = _xmp_from_pillow(img)`. Since `ALLOWED_IMAGE_EXTENSIONS` includes `heic`/`heif`, a phone photo uploaded as a profile picture, org logo, or cover image keeps its GPS coordinates. Reproduced against pillow-heif 1.5.0: the old `frombytes` path yields 0 EXIF tags on HEIF output, the new path preserves all of them (a JPEG `comment` marker also survives via the `im.info.get("comment")` fallback). The existing `test_strip_exif_removes_data` only covers JPEG, so it did not catch this. Suggested fix: keep the memory win but explicitly drop metadata before saving, e.g. `image.info = {}` (or `image.info.pop("exif", None)` / `"xmp"` / `"comment"`), and add a HEIF case to the test.

https://github.com/letsrevel/revel-backend/blob/b9f22409fc9306bbdc047c3d4255a9903381dbec/src/common/utils.py#L28-L34

https://github.com/letsrevel/revel-backend/blob/b9f22409fc9306bbdc047c3d4255a9903381dbec/src/common/fields.py#L39-L41

PR #919, opened 2026-09-02
feat(auth): generic OpenID Connect login

biagiodistefano:
## Summary

Generic OpenID Connect login (Revel as Relying Party). Any operator-configured issuer works — Google on letsrevel.io, Keycloak/Authentik/Zitadel/institutional IdPs for self-hosters — through one code path. Replaces the Google-only ID-token endpoint.

- **Flow:** backend-handled authorization code + PKCE. `GET /api/auth/oidc/{provider}/start` → IdP → `GET /api/auth/oidc/{provider}/callback` → 302 to `{FRONTEND_BASE_URL}/auth/callback?token=…` with a 60 s single-use JWT → `POST /api/auth/oidc/exchange` returns the normal access/refresh pair. The client secret and PKCE verifier never leave the backend; `state` is single-use in Redis **and** bound to the browser via an HttpOnly `SameSite=Lax` cookie on the API origin.
- **Verification:** JWKS by `kid` (`PyJWKClient`, re-keyed on `jwks_uri` change), alg allowlist RS256/ES256, `aud`/`iss`/`exp`/`iat`/`nonce`/`azp` checks, 30 s clock leeway, pydantic-validated claims.
- **Accounts:** new `ExternalIdentity(user, provider, subject, email)` table, unique on `(provider, subject)`, with an idempotent backfill from `django-google-sso`. Auto-link to an existing account only when the ID token asserts `email_verified`; guests are upgraded; bans and inactive accounts refused; `is_staff`/`is_superuser` never derived from claims; IdP emails lowercased.
- **Policy change:** passwords and identities coexist. The "Login via SSO" lockout and the password-reset / email-change blocks are removed. TOTP is skipped for IdP logins (the IdP's MFA is trusted).
- **API:** `GET/DELETE /api/account/identities[/{provider}]` (unlink refuses if it would strand the account); `/version` gains `sso_providers` (`key`, `name` only).
- **Config:** `OIDC_PROVIDERS=google,keycloak` + `OIDC_<KEY>_ISSUER|CLIENT_ID|CLIENT_SECRET[|NAME|SCOPES]`, parsed at settings load into frozen dataclasses. `django-google-sso` stays for the Django admin login page only.
- **Docs:** self-hosting guide, ADR-0016 (library choice: hand-rolled on PyJWT + httpx over allauth / mozilla-django-oidc / authlib), changelog, five translated catalogs.

## ⚠️ Breaking (`.env`)

`FEATURE_GOOGLE_SSO` is removed. Use `OIDC_PROVIDERS=google` with `OIDC_GOOGLE_ISSUER=https://accounts.google.com`, `OIDC_GOOGLE_CLIENT_ID`, `OIDC_GOOGLE_CLIENT_SECRET`. Register `https://<api-host>/api/auth/oidc/google/callback` at Google. `POST /api/auth/google/login` is gone and `features.google_sso` on `/version` is replaced by `sso_providers`. The frontend never used either, so deploying the backend first is safe.

## Deploy checklist

- [ ] infra repo: replace `FEATURE_GOOGLE_SSO` in `setup.sh` / `.env.example` with the `OIDC_*` vars
- [ ] Google Cloud console: OAuth consent screen "In production" (lifts the 100-user testing cap; no verification needed for `openid email profile`)
- [ ] Frontend PR: `/auth/callback` route, provider buttons from `sso_providers`, identities settings, `pnpm generate:api`

## Test plan

- [x] 706 tests in `src/accounts` + `src/api` (new: config parser, model + backfill, hand-off token, discovery/state/PKCE, code exchange + ID-token rejections incl. HS256 confusion, user resolution matrix, redemption single-use, identities, controller redirects and cookie binding, `/version`)
- [x] `test_error_response_contracts.py`, `test_openapi_schema.py`
- [x] `make check` equivalents on tracked files (ruff, mypy --strict, migration-check, i18n-check, file-length, task-names)
- [ ] Manual: local Keycloak realm with `OIDC_KEYCLOAK_ISSUER=http://localhost:8080/realms/dev` (http allowed only under `DEBUG`)

## Follow-ups (not blocking)

- Strict single-use on concurrent hand-off redemption (same pattern as impersonation) via `OutstandingToken` `created` flag
- Unknown provider on callback answers JSON 404 instead of a login-page redirect
- Shared one-shot-token helper for impersonation + OIDC hand-off

## Summary by CodeRabbit

* **New Features**
  * Added configurable OpenID Connect login for Google, Keycloak, Authentik, Zitadel, and other providers.
  * Added account settings to view and unlink connected sign-in identities.
  * Identity-provider accounts can set passwords and change email addresses.
  * The version endpoint lists configured sign-in providers without exposing credentials.

* **Changed**
  * Existing Google sign-in identities are preserved during migration to OIDC.
  * Replaced standalone Google sign-in configuration and endpoint with generic OIDC configuration.

* **Documentation**
  * Added OIDC setup guidance and architecture documentation.


biagiodistefano:
### Code review

Found 2 issues:

1. New accounts are created with `email_verified=True` even when the ID token asserts `email_verified: false`. The `claims.email_verified` gate at line 304 only guards the link-to-existing-account branch; the `else` branch that creates a fresh `RevelUser` hardcodes `"email_verified": True` regardless of the claim. Against any configured issuer that allows self-registration without address verification (the self-hosted Keycloak/Authentik audience this PR targets), an attacker who registers a not-yet-used victim address at the IdP gets a Revel account that passes `requires_verified_email` gates (e.g. org creation) and matches email-addressed invitations. This contradicts the trust model in [ADR-0016](https://github.com/letsrevel/revel-backend/blob/2fef3ecfc07589ecdb6dd9bbb49dd73bf2514f5c/docs/adr/0016-oidc-relying-party.md) ("only when the ID token asserts `email_verified: true`; otherwise refuse"), and `test_oidc_resolve_user.py` has no unverified-create case. (bug: `_resolve_user` create path vs. link path make opposite trust decisions on the same claim)

https://github.com/letsrevel/revel-backend/blob/2fef3ecfc07589ecdb6dd9bbb49dd73bf2514f5c/src/accounts/service/oidc.py#L303-L334

2. Identity-only accounts now get a misleading `400 "Incorrect password."` from `POST /account/email-change-request`. The PR removes the SSO pre-check that ran before `check_password`; that check was added in the PR #422 review-fix pass specifically so accounts with a sentinel/unusable password get a clear error instead of "Incorrect password" (inline comment at the time: "SSO check must run before the password check"). New OIDC accounts still call `set_unusable_password()` (`oidc.py:333`) and the backfilled legacy cohort has `password=""`, so `check_password` is always `False` for them and the endpoint fails with no hint that a password reset is the fix. The PR body says identity-provider accounts "can now ... change email addresses", but the only test for that path (`test_email_change_allowed_with_linked_identity`) uses a user that already has a password; the deleted `test_google_sso_user_rejected` covered the real sentinel-password case. A `has_usable_password()` guard with a "set a password first" message restores the intent. (bug due to removed guard in `request_email_change`; historical context in commit `2e5a9aac`)

https://github.com/letsrevel/revel-backend/blob/2fef3ecfc07589ecdb6dd9bbb49dd73bf2514f5c/src/accounts/service/account.py#L412-L416

PR #920, opened 2026-09-03
feat(e2e): Keycloak OIDC provider in the e2e compose stack

biagiodistefano:
## Summary

Adds a real OpenID Connect provider to the **e2e compose overlay only**, so the frontend Playwright suite can exercise the generic OIDC login shipped in #919 end to end.

- `docker-compose-e2e.yml`: `keycloak` service — Keycloak 26.7.3, `start-dev --import-realm`, health-checked on the management port (no curl in the image, so a bash `/dev/tcp` probe), published on `127.0.0.1:8080`, 768 MB / 1 CPU cap. Not in `compose.yaml`: `make run` / `make test` never start it.
- `e2e/keycloak/revel-realm.json` (+ `README.md`): realm `revel`, confidential client `revel-backend` (secret `e2e-secret`, redirect URI `http://localhost:8000/api/auth/oidc/keycloak/callback`, PKCE S256), `verifyEmail` off, three seeded users with `password123`:
  - `charlie.member@example.com` — mirrors the bootstrap persona → auto-link to the existing account, unlink allowed
  - `sso.only@example.com` — not a Revel user → first login creates the account, unlink refused until a password is set
  - `unverified.sso@example.com` — `emailVerified: false` → `/login?error=oidc_unverified_email`
- `Makefile`: `E2E_OIDC_ENV` (`OIDC_PROVIDERS=keycloak`, issuer `http://localhost:8080/realms/revel`, client id/secret) injected inline into the e2e gunicorn like the PgBouncer vars; `run-e2e`, `run-e2e-daemon` and `e2e-seed` now `--wait` on `keycloak` too. The http issuer is accepted because the dev `.env` has `DEBUG=True`.

Keycloak rejects unknown keys in the realm file (an initial `_comment` key failed the import), which is why the notes live in the README.

## Verified locally

1. Discovery served from the pinned issuer, `code_challenge_methods_supported` includes `S256`, both `client_secret_post` and `client_secret_basic` advertised.
2. `make run-e2e-daemon` → `GET /api/version` reports `sso_providers: [{"key": "keycloak", "name": "Keycloak (e2e)"}]`; features unchanged.
3. Browser: `/api/auth/oidc/keycloak/start?return_url=/events` → Keycloak login page (state, nonce, S256 challenge in the URL) → sign in as Charlie → backend callback → `http://localhost:5173/auth/callback?token=…` (frontend route pending in letsrevel/revel-frontend#875).
4. `POST /api/auth/oidc/exchange` with that token → Charlie's pair + `return_url: "/events"`; second exchange → 401; `GET /api/account/identities` lists the Keycloak link with Keycloak's `sub`.

Test row and daemon cleaned up afterwards; nothing in `make check`'s scope changed (no Python touched).

## Notes for reviewers

- `make e2e-setup` (frontend repo) now includes Keycloak's cold start (~20–30 s) via `--wait`.
- Per-run throwaway users can be created through Keycloak's admin REST API (`admin`/`admin`), documented on revel-frontend#875 together with the spec list.
- Nothing to reseed: personas come from `make bootstrap` as before; the realm is re-imported on every container start.

## Summary by CodeRabbit

* **New Features**
  * Added a Keycloak-based OpenID Connect provider to the end-to-end testing environment.
  * Added a preconfigured `revel` realm with a backend client and representative verified, SSO-only, and unverified test users.
  * The end-to-end stack now starts and health-checks Keycloak automatically.

* **Documentation**
  * Added setup and configuration guidance for the local Keycloak realm and its test accounts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,12 @@ run-e2e:
 # safety) — e2e-setup in revel-frontend calls this before run-e2e-daemon.
 .PHONY: e2e-seed
 e2e-seed:
-	docker compose -f compose.yaml -f docker-compose-e2e.yml up -d --wait pgbouncer keycloak
+	docker compose -f compose.yaml -f docker-compose-e2e.yml up -d --wait pgbouncer
+	# Keycloak keeps its realm in an embedded H2 DB inside the container and
+	# `--import-realm` skips a realm that already exists, so a reused container
+	# carries admin-API users and stale realm JSON across runs. Recreating it (no
+	# volume) is the Keycloak equivalent of reset_events below.
+	docker compose -f compose.yaml -f docker-compose-e2e.yml up -d --wait --force-recreate keycloak
 	uv run python src/manage.py reset_events --no-input
 	$(MAKE) seed
 	$(MAKE) bootstrap-tests
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,15 @@ run:
 # which SIGKILLs the workers. The var disables that guard; it's a no-op on Linux/CI.
 # Shared gunicorn invocation for run-e2e / run-e2e-daemon — see the comment on
 # run-e2e for why gthread+PgBouncer and the ObjC fork-safety var.
-E2E_GUNICORN = DB_USE_PGBOUNCER=True DB_PORT=6432 OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES uv run gunicorn revel.wsgi:application \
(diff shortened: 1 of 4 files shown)


biagiodistefano:
Addressed the realm-reset finding from both reviewers in the latest commit: `make e2e-seed` now `--force-recreate`s the `keycloak` container (no volume, so the H2 state is discarded and the realm re-imports), while `run-e2e`/`run-e2e-daemon` keep a plain `up` since `e2e-setup` runs the seed first. Verified by creating a user through the admin API, recreating, and listing users again: only the three seeded ones remain. README and compose comment corrected accordingly.

PR #924, opened 2026-09-04
feat(integrations): Eventbrite listing sync — platform listings via a provider protocol (phases 1–3)

biagiodistefano:
## Summary

Long-lived integration branch for external listing platforms (Eventbrite first). All three phases are in: **Phase 1 (foundation)**, **Phase 2 (sync)**, **Phase 3 (live data)**. Design: ADR-0017 (`docs/adr/0017-platform-listings-provider-protocol.md`). Design: `docs/superpowers/specs/2026-09-04-platform-listings-design.md` (local, gitignored).

### Phase 1 — what's here
- New `integrations` app: `ListingProvider` `t.Protocol` + neutral pydantic shapes; registry gated on `INTEGRATIONS_EVENTBRITE_CLIENT_ID/_SECRET`; models `PlatformConnection` (encrypted tokens), `EventLink`, `TierLink`, `WebhookDelivery` (one migration).
- Stable `IntegrationErrorCode` contract + per-app exception handler → `{detail, code, provider_message}`.
- Owner-only connection endpoints under `/organization-admin/{slug}/integrations` (list, connect, accounts, select-account, update auto-sync, disconnect); signed-state OAuth flow (state bound to org+user, cookie-matched, ownership re-checked at callback); public callback that always redirects; public webhook receiver that records deliveries (dispatch arrives in phase 3).
- Eventbrite provider: OAuth code exchange, account listing, webhook (un)registration, unsigned-webhook parser with host/port/path guard (SSRF).
- `events`: `TicketTier.sales_paused` organizer kill switch — enforced on the purchase path, eligibility gate (`sales_paused` reason), listings and schemas; one migration. Event duplication resets it.
- Contract tests, self-hosting docs (`docs/self-hosting/tiers.md`, `index.md`), translations (de/es/fr/it/pt, no fuzzy).

### Endpoints
| Method | Path | Auth |
|---|---|---|
| GET | `/organization-admin/{slug}/integrations` | owner |
| POST | `/organization-admin/{slug}/integrations/{provider}/connect` | owner |
| GET | `/organization-admin/{slug}/integrations/{provider}/accounts` | owner |
| POST | `/organization-admin/{slug}/integrations/{provider}/select-account` | owner |
| PATCH | `/organization-admin/{slug}/integrations/{provider}` | owner |
| DELETE | `/organization-admin/{slug}/integrations/{provider}` | owner |
| GET | `/integrations/{provider}/callback` | none (redirect only, hidden) |
| POST | `/integrations/{provider}/webhook/{secret}` | none (hidden) |

### Notes for reviewers
- Callback view contains a deliberate, documented try/except (browser mid-navigation → every failure must redirect; the redirect needs the org slug only the view can resolve).
- `PlatformConnection.provider` is intentionally unconstrained; `registry.get_provider()` validates at every entry point and the admin is read-only.
- Eventbrite facts from the spike: tokens never expire (no refresh), `state` round-trips, **rate limit 2000/h per token and per app key**, CloudFront 403s loopback redirect URIs (local dev needs a hosts alias).
- Frontend contract (for the FE issue): call `connect` from the browser with `credentials: 'include'`; landing page `/org/{slug}/admin/integrations?connected|select|error=<code>`.

### Phase 2 — sync (what's here)
- Provider protocol read/write half + neutral `RemoteEvent` / `RemoteTicketClass` / `RemoteVenue` shapes; `FakeProvider` in-memory remote store for tests.
- `service/mapper.py`: Revel → neutral, pure. Eligibility (`event_private` / `event_open_ended` / `event_no_tickets`), tier skip rules (`tier_variable_price`, `tier_members_only`, `tier_seated`, `tier_offline_payment`, `tier_no_capacity`, `tier_currency_mismatch`), 140-char summary, HTML body, venue with ISO country, `listed` follows Revel visibility.
- `service/sync_service.py`: `request_push` → Celery `integrations.push_event_link` (full-state push; tier reconciliation acts only on Revel-known classes — remote-only classes are left alone and reported as `remote_only_tier`; unsold stale → delete, sold → hide; class deleted remotely → recreated; status rule: first push = draft, live+CANCELLED → cancel, live+DRAFT → `unpublish_refused`; `remote_event_missing` → `broken`, next push recreates; retry only on transient errors with `provider_rate_limited` report entries; `pending` is never terminal). Explicit `publish_link`; per-link auto-sync override.
- `service/import_service.py`: `remote-events` picker + `import` → Celery `integrations.import_remote_event`; draft `Event` + tiers from remote data (HTML sanitized before Markdown, lengths truncated, invalid classes skipped with a report entry); link created last so auto-sync stays quiet; idempotent via a conditional unique constraint on `(connection, remote_id)`.
- `signals.py`: opt-in auto-sync on Event/TicketTier saves — only maintains listings already pushed once; 30 s debounce via `cache.add` (fail-open on cache errors); dispatched `on_commit`.
- Eventbrite: `translate.py` (the only module with Eventbrite field names) + full read/write methods (venue reuse on update, structured content for the long description, publish/cancel, ticket class upsert/delete/hide); tested on spike-recorded fixtures.

### Phase 2 endpoints
| Method | Path | Auth |
|---|---|---|
| GET | `/event-admin/{event_id}/integrations` (one row per enabled provider: `connection_status` + `link` or null) | staff `manage_event` |
| POST | `/event-admin/{event_id}/integrations/{provider}/push` (202) | staff `manage_event` |
| POST | `/event-admin/{event_id}/integrations/{provider}/publish` | staff `manage_event` |
| PATCH | `/event-admin/{event_id}/integrations/{provider}` (auto-sync override) | staff `manage_event` |
| GET | `/organization-admin/{slug}/integrations/{provider}/remote-events` | owner |
| POST | `/organization-admin/{slug}/integrations/{provider}/import` (202) | owner |

### Phase 2 notes for reviewers
- `0001_initial` was regenerated (still unmerged): `EventLink.remote_id` allows blank; `TierLink.tier` is `SET_NULL` so a deleted Revel tier leaves the link that reconciliation needs; conditional `UniqueConstraint(connection, remote_id)`.
- Product decision to confirm: synced events are created **listed** on Eventbrite when the Revel event is public (unlisted otherwise). The spike used `listed: false` only because it was a throwaway account.
- Known/deferred: no per-link mutual exclusion between a manual push and a debounced auto push (reconcile self-heals); `IntegrationErrorCode` will move to `integrations/enums.py` in phase 3 so schemas can use the model enums instead of `Literal` mirrors; import failures leave no organizer-visible trace yet (spec gap).

### Phase 3 — live data (what's here)
- Webhook deliveries dispatch `on_commit` to `integrations.handle_webhook_delivery`: the provider resolves the untrusted pointer (only `GET /orders/{numeric id}/` on its own host; event/attendee paths carry the id), the link is found by `(connection, remote_id)`, then `order_changed` → debounced count refresh (key released on transient failure; delivery stays `received` for the retry) or publish/unpublish → an explicit status matrix (`draft⇄live`; cancelled/broken never move). Unknown events → `ignored`; failures → `failed` + re-raise.
- Count refresh reuses the expanded `get_event` (no extra protocol method); `TierLink.remote_quantity_sold`/`counts_updated_at` per class.
- Remote pause/resume: `POST /event-admin/{event_id}/integrations/{provider}/pause|resume` (optional `{tier_id}`), synchronous, per-tier results; tiers not attempted after a revoked token are reported with `connection_revoked`; `remote_paused` is written only here and survives pushes.
- Beat: `integrations.reconcile_counts` every 15 min (stalest first via `Min`, never `.distinct()`; tierless links skipped; stops when the shared app-key budget parsed from `x-rate-limit` drops below `INTEGRATIONS_RATE_RESERVE`=200) and `integrations.prune_webhook_deliveries` daily 04:50 UTC (`INTEGRATIONS_WEBHOOK_DELIVERY_RETENTION_DAYS`=30). Rows registered by data migration `0002_beat_tasks`.
- `TicketTierDetailSchema.external_sales[]` (`provider`, `quantity_sold`, `updated_at`, `paused`) via the reverse relation — `events` still imports nothing from `integrations`; the FE sums per-tier for the event total.
- `IntegrationErrorCode` lives in `integrations/enums.py`; schemas use model enums (`ConnectionStatus`, `RemoteStatus`, `SyncState`, `Origin` OpenAPI components). New codes `tier_not_linked`, `stripe_not_connected`.
- Import + Stripe: paid Eventbrite classes become online tiers; if the organization is not Stripe-connected they start with `sales_paused` and the sync report carries `stripe_not_connected` — buyers never reach a checkout that would refuse, the organizer resumes after connecting. `ConnectionSchema.stripe_connected` lets the FE warn on the import picker. Free classes are unaffected.

### Phase 3 endpoints
| Method | Path | Auth |
|---|---|---|
| POST | `/event-admin/{event_id}/integrations/{provider}/pause` | staff `manage_event` |
| POST | `/event-admin/{event_id}/integrations/{provider}/resume` | staff `manage_event` |

### Migrations
- `integrations/0001_initial` (regenerated on this branch; never on main)
- `integrations/0002_beat_tasks` (data migration: two PeriodicTask rows)
- `events/0117_tickettier_sales_paused`

### Verification
`make check` green; `uv run pytest src/integrations src/events/tests/test_tier_sales_paused.py ...` 189 passed; branch coverage on `integrations` ≈ 98 %.

## Test plan (phase 3 additions)
- [ ] Place an order on Eventbrite → `external_sales.quantity_sold` updates within seconds (webhook) and the delivery shows `processed`
- [ ] Pause a tier from Revel → hidden on Eventbrite checkout; resume → visible; push in between does not unhide it
- [ ] Publish on Eventbrite directly → link shows `live`; unpublish → `draft`
- [ ] Watch the beat reconcile log for `integration_reconcile_done` and the budget warnings

## Test plan (phase 2 additions)
- [ ] Push a public event → Eventbrite draft with the mapped tiers; publish from Revel → live
- [ ] Enable auto-sync, rename the event → Eventbrite updates within ~30 s, venue not duplicated
- [ ] Add a ticket class directly on Eventbrite → survives the next push, shows as `remote_only_tier`
- [ ] Import an existing Eventbrite event → Revel draft with tiers, no push triggered

## Test plan (phase 1)
- [ ] Register an Eventbrite app for staging, set the two env vars, click Connect → approve → land on the settings page with `?connected=eventbrite`
- [ ] Multi-organization Eventbrite user gets the account picker
- [ ] Disconnect removes the webhook on Eventbrite
- [ ] Pause a tier → it disappears from checkout and a direct purchase POST is refused


biagiodistefano:
@coderabbit full review

biagiodistefano:
### Code review

Found 1 issue:

1. Retryable provider failures during webhook **order** resolution are marked `FAILED` instead of retried, so the order's count update is lost.

`EventbriteProvider.resolve_notification` makes a live `GET /orders/{id}/` call. `client.py` raises a plain `ProviderError(..., retryable=True)` on 429/5xx — it never raises the `RetryableProviderError` subclass (the only raise sites are `sync_service.py` L146 and L291). Since Python matches `except` by `isinstance`, the `except RetryableProviderError` guard below cannot catch it; execution falls into `except ProviderError`, the delivery is marked `FAILED`, and `handle_webhook_delivery` (declared `autoretry_for=(RetryableProviderError,)`) never retries. The inline comment states the opposite intent.

https://github.com/letsrevel/revel-backend/blob/1dd9acb711d5445856ea20030fbaf6920d36ef56/src/integrations/service/webhook_service.py#L117-L124

The order-shaped path that triggers the live call:

https://github.com/letsrevel/revel-backend/blob/1dd9acb711d5445856ea20030fbaf6920d36ef56/src/integrations/providers/eventbrite/provider.py#L154-L159

The sibling counts path does perform the upgrade this one is missing, which is why `webhook_service.py` L84 works and L118 does not:

https://github.com/letsrevel/revel-backend/blob/1dd9acb711d5445856ea20030fbaf6920d36ef56/src/integrations/service/sync_service.py#L143-L147

Suggested fix: in the `except ProviderError as e:` branch, re-raise as `RetryableProviderError` when `e.retryable`, mirroring `refresh_counts`.

Reachable on any `order.placed` / `order.updated` webhook when the shared app-key budget is exhausted — i.e. during high-volume on-sales. Not covered by tests: `test_retryable_failure_releases_debounce_key_and_leaves_delivery_received` uses an event-shaped `api_url` (counts path), and the `resolve_notification` failure tests use non-retryable codes only.

biagiodistefano:
### Code review — Copilot findings validated, and fixes pushed

I independently re-verified each of Copilot's 12 inline findings against the code and scored them 0-100 for confidence. **Zero were false positives**, which is a good hit rate; most are real but individually survivable, because the system converges (the next push re-reconciles, the 15-minute sweep re-refreshes).

| Finding | File | Score | Fixed |
|---|---|---|---|
| Retryable error not upgraded → delivery `FAILED`, never retried | `webhook_service.py:124` | 100 | yes |
| Zero-decimal currencies (JPY/KRW) listed at 100x price | `translate.py:75` | 75 | yes |
| Budget cache write not fail-open → lost remote ID → duplicate listing | `client.py:44` | 75 | yes |
| `list_accounts` ignores pagination | `provider.py:106` | 75 | yes |
| `venue_id` never cleared on update → stale remote venue | `provider.py:240` | 75 | yes |
| Free-tier import falls back to instance default currency | `import_service.py:82` | 75 | yes |
| Majority currency counts tiers that will be skipped | `mapper.py:168` | 72 | yes |
| Retry exhaustion leaves the delivery `RECEIVED` forever | `tasks.py:93` | 72 | yes |
| `resume` forces `hidden=False`, exposing a locally-paused tier | `sync_service.py:443` | 68 | yes |
| Structured-content body not read on import → empty description | `provider.py:192` | 60 | yes |
| All retryable failures reported as `provider_rate_limited` | `sync_service.py:314` | 52 | yes |
| Import doesn't validate `remote_id` against the selected account | `import_service.py:105` | 25 | no — see below |

Three of these are one root cause: an error path that loses the information needed to recover. `client.py` signals transience with a `retryable=True` **flag**, but `webhook_service` and the tasks branch on the `RetryableProviderError` **type**, and only `sync_service.refresh_counts` bridged the two. That is why the guard at `webhook_service.py:118` could never fire for `resolve_notification`, and why the report code collapsed to `provider_rate_limited`.

Notable behaviour changes worth a look:

- `mapper.tier_hidden()` is now the single source of the hide rule, read by both the push and the remote resume — so clearing an organizer's remote pause can no longer un-hide a tier that is unlisted or locally `sales_paused`.
- `get_description()` is a new optional read on the provider protocol, deliberately kept **off** `get_event()`: that call doubles as the counts refresh on every order webhook and on the 15-minute reconcile, and must stay one request.
- A paused tier now 403s with "Ticket sales are paused." instead of "You're outside of the sale window.", and no longer shows `can_purchase: true` to anonymous viewers. Both reuse an existing msgid, so no catalog churn.

Two things I did **not** change, deliberately:

1. **Import account validation** (scored 25). Each Revel org holds its own OAuth token, so there is no cross-tenant path — the worst case is an organizer importing from another of their *own* Eventbrite orgs, which they already have legitimate access to. The asymmetry is still real (`list_events` is account-scoped, `get_event` is not), but tightening it could break a legitimate multi-org workflow, so it is your call rather than mine.
2. **The `Capabilities` dataclass** is declared and assigned but never read anywhere; its docstring says the shared mapper uses it to "warn instead of crashing", and `mapper.py` instead hardcodes the Eventbrite assumptions. That weakens ADR-0017's "a second platform is a client + translator ... no changes to the sync service". Either wire it up or drop it — but that is a design call, not a bug fix.

Also still open: `anonymous_can_purchase()` does not check the sales *window* at all. That gap predates this PR, so I only added the `sales_paused` half and left a comment pointing at it.

Verification: `ruff format`, `ruff check`, `mypy --strict`, `i18n-check`, `task-names` and `file-length` all clean; 321 integrations + sales-paused + error-contract tests pass, plus 236 permission/public-listing/guest-checkout tests. New regression tests cover every fix above.

biagiodistefano:
### Verified against the live API — one correction, two decisions

I checked the remaining uncertain claims against real Eventbrite on the spike account (throwaway drafts, all deleted afterwards). One of my earlier statements was wrong.

**Correction — the venue fix was a no-op, and the finding was mostly a false positive.**

I previously said `venue_id: ""` detaches a remote venue. It does not. Eventbrite accepts and silently ignores it, and `venue_id: null` too:

```
venue_id=null                       before='299309199'  after='299309199'  -> still attached
online_event=true                   before='299309200'  after=None         -> DETACHED
venue_id=null + online_event=true   before='299309203'  after=None         -> DETACHED
```

Only `online_event: true` detaches a venue — and `is_virtual` already sends that, so the scenario Copilot described (a physical event becoming virtual/locationless) **was already correct before any of my changes**, confirmed end-to-end:

```
venue after create        : 299309216
venue after virtual update: None | online: True
```

So that finding drops from 75 to roughly 25. I reverted the `venue_id: ""` payload — shipping a misleading no-op is worse than shipping nothing — and replaced the test with one pinning the behaviour that actually holds. The genuine residual is narrow: a *physical* event whose address and city are both cleared keeps its old remote venue, because Eventbrite has no representation for that state and detaching would mean recreating the listing, losing its URL and its orders. That is now recorded in ADR-0017 rather than papered over.

**Two other fixes, confirmed real by the same method.** JPY sends as `JPY,1000` and displays `¥1,000`, while EUR 15.00 sends `value=1500` — Eventbrite treats zero-decimal currencies exactly as Stripe does, so the old `×100` really did list a ¥1,000 tier at ¥100,000. And `description.html` comes back `None` even with content set, while structured content holds the body — so without the new read path an import genuinely lands with an empty description.

**Decision: import now honours the selected account.** `get_event` is not account-scoped the way `list_events` is, so a hand-crafted remote id could link — and later mutate — a listing in another account the same token reaches. The owning account rides along on `RemoteEvent` from the response Eventbrite already returns, so the check costs no extra call, and providers that do not report it are trusted as before. Not a cross-tenant hole (each org holds its own token), but the account picker is a boundary the organizer chose, and it should mean something.

**Decision: the `Capabilities` dataclass is deleted.** Nothing ever read it, and its docstring claimed the shared mapper warned from it — an unread flag table reads as enforcement that is not there, which is worse than no table. The flags a second provider actually needs are not knowable before it exists. ADR-0017 now states plainly which Eventbrite assumptions the mapper encodes (one currency per event, end time and capacity required, unpublish refused with orders) and that a differing platform will need those made conditional.

Still open and unchanged: `anonymous_can_purchase()` does not check the sales *window*. That predates this PR; only the `sales_paused` half is fixed, with a comment marking the gap.

Pushed in `9874f189`. Full gate green — format, lint, `mypy --strict`, i18n-check, task-names, file-length — with 323 tests passing.

PR #973, opened 2026-09-10
fix: resolve the 22 medium findings of the 2026-09-10 tech-debt sweep (#961–#972)

biagiodistefano:
Integration branch for the 22 medium-priority findings (7–28) of `2026-09-10-TECH-DEBT_REPORT.md`. Every finding was independently re-verified against HEAD before being fixed (20 confirmed; 23 and 25 narrowed to a corrected scope). The twelve fixes were reviewed and merged as separate PRs into this branch so each keeps its own squash commit:

| PR | Findings | Commit |
|---|---|---|
| #961 | 14 — `configure_stripe()` replaces 11 copy-pasted init triples | refactor(stripe) |
| #962 | 24, 25, 26 — template/telegram coverage, fixture hoist, five test-file splits | test |
| #963 | 7, 11, 18, 27 — drop 7 unused deps, declare pyjwt/jinja2, delete playground.py, pydantic-mypy `init_typed` | chore(deps) |
| #964 | 8, 9, 28 — Telegram `/preferences` dead surface, phantom settings, legal URLs from SiteSettings, FSM TTL | refactor(telegram) |
| #965 | 10 — legacy `#2196F3` → brand token; shared banned-hex list | fix(notifications) |
| #966 | 12, 21 — retake/staleness and applicability policy helpers | refactor(questionnaires) |
| #967 | 15, 20 — wallet 503 handlers with `code`, shared pass field resolution | feat(wallet) |
| #968 | 16 — 76 user-facing error strings translated + `make i18n-literals` guard | feat(i18n) |
| #969 | 17 — drop unusable btree index on request `message`, 500-char cap | fix(events) |
| #970 | 22 — `notify_org_staff()`; staff preferences honoured on 7 more fan-outs | refactor(notifications) |
| #971 | 23 — upload workflow out of `common/utils.py`, import cycle gone | refactor(common) |
| #972 | 13 — `events/service/subscription/` package | refactor(subscriptions) |

## Behaviour changes to be aware of
- **#966**: a retake is allowed at exactly `retry_on`; a zero `max_submission_age` means no expiry. NULL `can_retake_after` semantics are now documented on the model and in both architecture docs (they differ by design between admission and membership).
- **#970**: seven more staff notification types honour preferences. Three of them (membership, invitation, whitelist requests) announce inbound work; if those should never be silenceable, that belongs at the preference layer.
- **#964**: unknown bot commands now fall through silently instead of replying "administrators only"; `/start` sends `ReplyKeyboardRemove` to clear the stale persisted keyboard.

## API contract changes (frontend: letsrevel/revel-frontend#916)
- `EventInvitationRequestCreateSchema.message` gains `maxLength: 500`.
- Eight wallet routes declare `503: WalletPassErrorSchema | ErrorDetail`; the transient-failure body carries `code: "wallet_pass_unavailable"`, the "not configured" body carries no code.

## Deploy notes
- Deploy this backend **before** switching Redis to `maxmemory-policy volatile-lru` in infra (finding 19; infra edits are staged locally, uncommitted), so aiogram FSM keys carry a TTL first.
- New env var `AIOGRAM_FSM_TTL_SECONDS` (default 86400). `TELEGRAM_SUPERUSER_IDS` / `TELEGRAM_STAFF_IDS` are gone (they were never read).
- Two migrations: `events/0119_drop_userrequest_message_index`, `questionnaires/0013_alter_can_retake_after_help_text`.

## Verification
Each PR passed CI on its own; the last two were re-run after rebasing onto this branch. On an integrated checkout equivalent to this branch tip: lint, mypy (1339 files), file-length, task-names, i18n-literals, migration-check and i18n-check all green; 9020 tests passed across events, questionnaires, wallet, telegram, notifications, common, accounts, api and polls.


biagiodistefano:
### Code review

Found 2 issues:

1. The test-file split dropped the body of `test_eu_buyer_zero_vat_invoice_is_not_export`. On `main` the test ends with `assert _is_export(self._invoice(buyer_vat_country="DE")) is False`; in the split file it is now only a docstring, so it passes without checking anything. The per-file `def test_` counts still match, which is why the split's self-check did not catch it. (bug due to `src/events/tests/test_attendee_invoice/test_generation.py` ending after the docstring; compare `src/events/tests/test_attendee_invoice_service.py` lines 573–575 on `main`)

https://github.com/letsrevel/revel-backend/blob/ca738012c98d14ffb381b91cb365308030c4a6fc/src/events/tests/test_attendee_invoice/test_generation.py#L455-L458

2. The new `configure_stripe()` guard does not cover the whole `subscription/stripe/` package it was written for. `STRIPE_CONFIGURING_MODULES` lists `base`, `payloads`, `checkout` and `sync` but omits `subscription.stripe.fees` (which does call `configure_stripe()`, just untested) and `subscription.stripe.plan_change`, which makes its own `stripe.Subscription.*` / `stripe.SubscriptionSchedule.*` calls and never calls `configure_stripe()` at all, relying on its `base`/`payloads` imports' side effects. That is exactly the dependency the helper's docstring says no module may have; the omission was pre-existing, but the centralizing commit and its "every module" test are where it should have been closed. Add the call to `plan_change.py` and both modules to the list. (bug due to `src/events/service/subscription/stripe/plan_change.py` having no `configure_stripe()` call while `src/common/service/stripe_config.py` states "Called at import time by every module that makes an outbound stripe-python call")

https://github.com/letsrevel/revel-backend/blob/ca738012c98d14ffb381b91cb365308030c4a6fc/src/events/tests/test_service/test_stripe_http_client_config.py#L25-L39

PR #990, opened 2026-09-18
feat(oauth): OAuth 2.1 / OpenID Connect provider — scoped third-party apps, MCP hosts, Sign in with Revel

### LINKED ISSUE #986: feat(oauth): OAuth 2.1 / OpenID Connect provider — scoped third-party apps, MCP hosts, Sign in with Revel
## Context

Today the API is reachable only with a Revel session JWT. There is no way for an organizer to connect a third-party tool (a Zapier-style connector, their own script, or an MCP host such as Claude Desktop) and no "Sign in with Revel" for third-party sites. ADR-0016 built the OIDC *relying party* side and explicitly left the *provider* role out of scope.

This issue makes Revel an **OAuth 2.1 authorization server and OpenID Provider**: a user approves a consent screen once, and the app calls the API with a token limited to the granted scopes **and** to what that user may already do through the existing org permission model.

Brainstormed and independently reviewed on 2026-09-14/15. The full design spec and the implementation plan live locally under `docs/superpowers/` (gitignored); the decisions below are the contract.

## Decisions (final)

| Question | Decision |
|---|---|
| Actor model | **User-delegated.** Effective power = granted scopes ∩ the user's current org permissions. App-as-its-own-principal (org installation) is a documented upgrade path, not v1. |
| Provider scope | **Full OIDC Provider** from day one: `openid`, `id_token` (RS256), `userinfo`, JWKS, both discovery documents. |
| Client registration | **Self-service developer portal** in account settings **plus RFC 7591/7592 dynamic client registration** (MCP hosts register themselves). Admin `verified` flag removes the consent-screen warning. |
| Scopes | **Coarse public scopes resolved to `PermissionKey`s** in a code registry: `openid profile email offline_access me:read me:rsvp org:read org:events org:tickets org:checkin org:members org:announcements org:questionnaires org:polls org:potluck`. `edit_organization`, purchases and owner-only routes are unreachable by app tokens. |
| Library | **django-oauth-toolkit 3.4.x.** It ships RFC 8707 resource indicators, RFC 9728 protected-resource metadata and DCR, all three required by the MCP authorization profile; Authlib lacks the first two. Evidence table goes into ADR-0018. |
| Tokens | Opaque access tokens, hashed at rest, 1 h; refresh 30 d with rotation and reuse detection (token families), no grace window; `offline_access` actually gates the refresh token. |
| Consent | **Headless**: backend validates/issues via DOT's `OAuthLibCore`, the SvelteKit frontend renders `/oauth/authorize`. DOT's session-based authorize view is not mounted. |
| Feature flag | Credential presence (ADR-0008): enabled iff `OIDC_SIGNING_KEY_PATH` is set. |

## Scope (backend)

1. **`oauth` app** with a swapped `OAuthApplication(AbstractApplication)` (owner, description, logo, URLs, `allowed_scopes`, `verified`, `is_active`, `last_used_at`; `is_usable()` = `is_active`). DOT's token models stay stock. One schema migration with `run_before = [("oauth2_provider", "0001_initial")]` (documented hand-edit exception, guarded by a test).
2. **Scope registry** (`oauth/scopes.py`) and a DOT scopes backend that is the single place per-app scope restriction lives.
3. **`ScopedJWTAuth`** in `common/authentication.py`: session JWT first, then DOT `verify_request`; refuses `user=None` tokens, DCR registration tokens, inactive users; re-applies `requires_verified_email`; binds `oauth_client_id` into structlog. Controllers opt in; anything left on `I18nJWTAuth` rejects app tokens.
4. **Enforcement**: `scope_allows()` as the first line of the six keyed permission classes (before the owner and potluck-creator short-circuits); `RequireScope("...")` for routes with no key; a convention test that fails for any `ScopedJWTAuth` route without a scope source, honouring ninja-extra's route-list-replaces-controller-list rule. 403 `insufficient_scope` and 401 `invalid_token` carry RFC 6750/9728 `WWW-Authenticate` headers.
5. **Protocol endpoints** mounted from DOT under the `oauth2_provider` namespace (its views `reverse()` those names): `/.well-known/openid-configuration`, `/.well-known/oauth-authorization-server`, `/.well-known/oauth-protected-resource`, `/o/jwks`, `/o/token`, `/o/revoke`, `/o/userinfo`, `/o/register[/{client_id}]`. Metadata views subclassed so `authorization_endpoint` points at the frontend. Throttled via a small decorator; DCR has a per-IP throttle and a global daily cap. **No introspection endpoint** in v1.
6. **Headless consent** `GET/POST /api/oauth/authorize` (session JWT): `prompt` handling, auto-approve on prior grant, `interaction_required` for `prompt=none`.
7. **Developer portal** `/api/oauth/apps` (verified email required): CRUD, secret shown once, rotate, logo, deactivate/activate, per-user cap, scope-shrink revokes tokens, `connections_count` but never the users.
8. **Connected apps** `/api/oauth/connections`: list + revoke, derived from DOT's tables.
9. **Validator**: claims per scope (`sub`, `name`, `preferred_username`, `picture`, `locale`, `updated_at`, `email`, `email_verified`), never staff/superuser/groups; refresh token dropped without `offline_access`.
10. **Lifecycle**: `blacklist_user_tokens` also revokes DOT tokens (lazy import); admin `save_model` and the API deactivate path revoke; two daily beat tasks (`oauth.clear_expired_tokens`, `oauth.prune_unused_dynamic_clients`, the latter ignoring DCR's eternal registration token).
11. **Admin** (Unfold) for apps and DOT token models + sidebar group; `/version` gains `features.oauth_provider`.
12. **Config**: `revel/settings/oauth.py`; `generate_oidc_signing_key` command; `.env.example`.
13. **Docs**: ADR-0018, `docs/developer-guide/oauth.md` (register an app, scopes, `curl` PKCE, connect an MCP host), engineering-notes entries, changelog, translations for all catalogues.

## Out of scope (decided)

MCP server implementation (it must later live under the same origin as `OAUTH_ISSUER` or issue unrestricted tokens); org-installation actors; per-organization grant restriction at consent; a separate `org:attendees` scope (attendee routes are keyed on `manage_event`/`manage_tickets`, so the labels of `org:events`/`org:tickets` say attendee details are included); introspection; client-credentials grant; RP-initiated logout; device flow; custom-scheme redirect URIs; purchase scopes; `edit_organization` scope; per-client throttling; `OptionalAuth` accepting app tokens; app marketplace.

## Testing

Full authorization-code + PKCE flow through the Django test client (happy path, wrong verifier, code reuse, refresh rotation and replay → family revoked, no `offline_access` → no refresh token, foreign `resource` → 401 on the API, deny → `access_denied`, deactivated app cannot refresh or hit userinfo, `prompt=none` both ways); parametrized scope enforcement over the registry; the default-deny convention test; claims incl. the staff-flag negative; discovery golden snapshots and 404 when disabled; DCR (public/confidential, loopback any port, custom scheme rejected, no owner even when logged in, RFC 7592 round trip, throttle and cap); migration ordering; lifecycle; error-response contract tests extended. Tests enable the provider through a session-scoped RSA key fixture and DOT's `setting_changed` reload.

## Acceptance criteria

- [ ] All of Scope 1–13 shipped on `feature/<this-issue>-oauth-provider`, `make check` green, coverage ≥ 90 % on `oauth/`.
- [ ] An MCP-style client (MCP Inspector or the `curl` walkthrough from the developer guide) completes discovery → dynamic registration → login → consent → PKCE exchange and calls `GET /api/organizations/` with the token.
- [ ] A token with only `org:read` gets 403 `insufficient_scope` on every write route; a token with `org:tickets` still gets 403 in an org where the user lacks `manage_tickets`.
- [ ] `id_token`/userinfo never contain `is_staff`, `is_superuser` or groups.
- [ ] With `OIDC_SIGNING_KEY_PATH` unset, every provider route returns 404 and nothing else in the API changes.
- [ ] User revoke, developer/admin deactivate, and global ban each invalidate live tokens at the next request, including refresh and userinfo.
- [ ] `.artifacts/openapi.json` regenerated.
- [ ] **On PR completion, open the follow-up issues:**
  - **revel-frontend**: `/oauth/authorize` consent page (app card, grouped scope list, unverified/auto-registered notices, allow/deny → `redirect_to`), Settings → "Connected apps", Settings → "Developer apps" (secret-once, logo, activate/deactivate), hide both when `/version.features.oauth_provider` is false, `pnpm generate:api`.
  - **infra**: mount `oidc.pem` beside the wallet certs (chmod 644 for uid 997), route `/.well-known/*` and `/o/*` to the API in Caddy, cache the discovery documents ≤ 1 h, add `OIDC_SIGNING_KEY_PATH` / `OAUTH_ISSUER` to the compose env, and extend the **self-deploy wizard** to prompt for the issuer and generate the signing key.
  - **documentation**: public docs page for third-party developers derived from `docs/developer-guide/oauth.md`, and a "What's New" entry.
- [ ] ADR-0018 merged with the library evidence table; engineering notes carry the `run_before` exception and the two-token-world revocation coupling.

biagiodistefano:
Closes #986.

Makes Revel an **OAuth 2.1 authorization server and OpenID Provider**, so a third-party app or an
MCP host can call the API on a user's behalf with scoped, revocable tokens. Built on
django-oauth-toolkit 3.4.1 (ADR-0018 records why, over Authlib).

Enabled iff `OIDC_SIGNING_KEY_PATH` is set (credential presence, per ADR-0008). `/version` reports
`features.oauth_provider`.

## The security model

An app token's effective power is **granted scopes ∩ the user's current org permissions**. An app
can never gain authority the user lacks — the scope only narrows. 14 coarse public scopes resolve
to the 23 internal `PermissionKey`s in a code registry; `edit_organization` maps to no scope, so no
app token can rename or delete an organization. `ScopedJWTAuth` tries the session JWT first and
only then hands the bearer to DOT, so **anything left on `I18nJWTAuth` or `OptionalAuth` rejects
app tokens by construction** — 33 controllers / 259 routes opt in, behind six mechanical guard
tests that assert that set, that every switched route enforces a scope, and that no owner-only or
read-scope-gated write is reachable.

## What's here

Swapped `OAuthApplication` (UUID pk) · scope registry · `ScopedJWTAuth`/`OAuthPrincipal` · scope
enforcement at seven keyed permission sites + `RequireScope` · OIDC validator with a 6-key claim
allow-list and a real `offline_access` gate · discovery, JWKS, token, revoke, userinfo and RFC
7591/7592 dynamic registration with IP-keyed throttling and a daily registration cap · headless
consent bound by a signed ticket · developer portal and connected-apps management · lifecycle
revocation, two daily beat sweeps, a readonly-by-construction credential admin · ADR-0018, a
developer guide, engineering notes, and 39 new strings translated into all five locales.

`make check` fully green. ~1250 tests in `src/oauth`; the pre-existing suites
(`events` 2165 + 574, `polls` 329, `questionnaires` 276, `accounts` 840, `common`, `api`) all pass
unchanged.

## Decisions I'd like you to review

Each of these was a judgement call made during implementation, with the reasoning and the
cost-if-wrong recorded.

1. **`org:tickets` carries refund authority.** `manage_tickets` gates real Stripe refunds and
   per-event revenue, so the public scope necessarily does too. The consent label now says so
   explicitly. It took three successive audits to find the full gate list — the plan, the scan and
   two reviewers each found less than the next. *If you disagree, the fix is splitting
   `manage_tickets` into finer keys (follow-up below), not a change here.*
2. **`org:read` does two jobs**, and that is the honest architectural weakness of this branch. It
   is both the admission ticket onto the organizer surface (on 232 routes purely as a gate) and a
   real read grant whose label promises specific data. Conflating them is the root cause of four
   separate findings during review — revenue/VAT, a Stripe billing-portal URL, the org-detail read
   with `vat_id`/`stripe_account_id`, and the member roster with emails and phone numbers. All four
   are now session-only, but a future `org:connect`-style baseline would make each permission list
   express one idea.
3. **Consent is bound by a signed ticket.** The backend originally held no proof the user saw the
   scopes it was granting — that property was enforced entirely by the frontend. `describe()` now
   returns a short-lived `consent_ticket` fingerprinting (user, client, redirect URI, displayed
   scopes, PKCE challenge **and method**, resource set); `decide()` requires it. Added now because
   the frontend does not exist yet, so the contract change is free.
4. **A global ban now deactivates the apps the banned user owns**, so every user of such an app
   loses access. Deliberate — a ban that leaves the banned party operating a client against other
   people's data has not ended their access. Reversible via `is_active`.
5. **Financial and money-movement surfaces are session-only.** Revenue, VAT, Stripe connect/verify,
   staff grant/permissions, and eleven attendee writes including the Stripe customer portal. This
   leaves an accounting integration with no app-token path at all (follow-up below) — the
   default-deny direction, chosen deliberately.
6. **GDPR export**: owned apps are exported minus `client_secret`; the DOT token tables are
   excluded as credential material, documented. **Art. 15 covers recipients of personal data, so a
   derived connected-apps summary is arguably the stronger answer — that is a policy call for you**
   (follow-up below).

## Exactly what "provider disabled" means

With `OIDC_SIGNING_KEY_PATH` unset, all 13 `/api/oauth/...` routes declare the gate and **none is
reachable**; no route answers 200 and none performs work. Stated precisely, because the issue's
acceptance criterion says "every provider route returns 404" and that is not literally achievable:

- **9 bodyless routes answer 404 unconditionally** (asserted by a registry-derived test).
- **4 routes carry a required body** — `POST /apps/`, `PATCH /apps/{id}`, `POST /apps/{id}/logo`,
  `POST /oauth/authorize`. Ninja validates the body *before* ninja-extra runs permissions, so a
  **well-formed** request to these answers 404 like the rest, while a **missing/empty** body
  answers 422 and **malformed JSON** answers 400 — refused on the body before the flag is
  consulted. A wrong body is refused whether or not the provider exists, so this leaks nothing
  beyond "this path parses bodies".
- Totality is carried by a **declaration** property (13/13 routes must declare `ProviderEnabled`,
  whatever their payload); observable behaviour by the 9-route status property. The route lists are
  derived from the API registry, not hand-written — a hand-written list is what let these two
  controllers ship ungated in the first place.

## User-visible changes

- **Changing your email address now disconnects every connected app.** Rotating the address
  invalidates every live session, and an app token is a live session.
- **A global ban deactivates the banned user's own OAuth apps** (see above).
- Discovery now advertises **`S256` only** for `code_challenge_methods_supported`, and
  `scopes_supported` no longer lists `me:rsvp`.
- `/o/userinfo` and `/o/register/{client_id}` are IP-keyed for authenticated callers too, so many
  users behind one NAT egress share the 60/min bucket.
- The OIDC `name` claim is **omitted** for a user with no preferred or full name — a `profile`
  grant can legitimately return `{"sub": …}` alone. Not an error.
- 259 operations move from `I18nJWTAuth` to `ScopedJWTAuth` in the OpenAPI spec. Both are
  byte-identical bearer schemes, so client behaviour is unchanged; the spec diff is large.

## Testing notes

Every behavioural change on this branch required **RED evidence against the broken code** before
the fix, and every new guard had to be shown to fail *alone* on its own hole. That was not
ceremony — the defects clustered in tests rather than code, and each of these was green CI:

- a guard that could never fire (the DCR scope was filtered out before the check saw it);
- a test that passed against a *valid* token while claiming to test an expired one
  (`RefreshToken.access_token` is a property that mints a fresh token per access);
- an `xfail` whose target could never satisfy it, on an `OptionalAuth` route;
- an assertion with no implementation behind it (DCR apps would have had zero scopes);
- a tautological check (`assert token not in archive`, where tokens are stored as SHA-256 only);
- **a negative security test that passed against a fully vulnerable admin, twice** — once from an
  unsplit `SplitDateTimeField`, once from a naive-datetime re-read that expired the forged row.
  A body the form rejects makes "the admin could not change X" pass for the wrong reason. Negative
  tests must assert the request was *accepted* before asserting the negative; this one now checks
  for a 302 first. Recorded in `docs/engineering-notes.md`.

## Follow-ups (not in this PR)

Required by the issue: **revel-frontend** (consent page, Connected apps, Developer apps — needs
`pnpm generate:api`; the consent page must forward the client's query string **verbatim**, since
altering a *value* breaks the ticket); **infra** (mount `oidc.pem` chmod 644 for uid 997, route
`/.well-known/*` and `/o/*`, cache discovery ≤1h, add the two env vars, extend the self-deploy
wizard); **documentation** (public developer page + What's New).

Recommended, arising from review: `org:financials` scope (5 has no path today) · `me:write` scope ·
**repo-wide `null`-on-PATCH sweep** — `Model.clean_fields()` skips validation for any `blank=True`
field whose value is in `empty_values`, so `X | None` + `exclude_unset` over a
`blank=True, null=False` column is a 500 on malformed input, and the pattern is not oauth-specific ·
`make lint` is blind to **every** package `__init__.py` (`[tool.ruff] exclude` globs `_*.py`; 16
pre-existing errors) · split `manage_tickets` · GDPR connected-apps summary · redirect-based
authorize errors (RFC 6749 §4.1.2.1) · cache `/o/jwks`.


biagiodistefano:
### Code review

Found 3 issues:

1. `POST /event-admin/{event_id}/actions/update-status/cancelled` with `refund_tickets=true` issues Stripe refunds for every online ticket on the event, but the route is gated by `manage_event`, which maps only to `org:events` ("Create, edit and delete events and event series, send invitations, and see attendee lists"). The PR's own rule (body item 1, commits 3904a9d4 / 46b24049 / c8fc756e) is that refund authority lives behind `org:tickets`, whose label was widened to say "issue refunds ... and see attendee details and revenue"; the single-ticket refund route in `EventAdminTicketsController` honours that, this bulk path does not. An app consented to `openid org:read org:events` can cancel an event and refund its whole ticket revenue against the organization's connected Stripe account. None of the six guards in `test_scope_coverage.py` sees it (property 5 only fires when the sole gate is a `:read` scope). Same controller, read side: `GET .../cancellation-refund-preview` returns `available_balance` from `stripe.Balance.retrieve` on the same `manage_event` gate.

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/src/events/controllers/event_admin/core.py#L124-L130

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/src/events/controllers/event_admin/core.py#L164-L172

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/src/oauth/scopes.py#L55-L69

2. `org:members` ("Manage members and subscriptions") unlocks `manage_subscriptions`, and the whole `OrganizationAdminSubscriptionsController` was switched onto `ScopedJWTAuth` on that key. On `org:read org:members` an app can read `GET /subscriptions/metrics` (MRR, churn, status breakdown), the org-wide payment ledger at `GET /subscriptions/payments` (amounts, member emails, Stripe invoice/payment-intent/refund ids), and `POST /payments/{id}/refund`, which flips a payment to `REFUNDED` and cancels the subscription when the refund covers the period. The PR body lists revenue as session-only and widened the `org:tickets` label for exactly this reason; the subscription surface got neither treatment.

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/src/events/controllers/organization_admin/subscriptions.py#L32-L40

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/src/oauth/scopes.py#L84-L86

3. `generate_oidc_signing_key` writes the PEM `chmod 0600`, and `settings/oauth.py` reads it with a bare `Path(path).read_text()` at settings import. `.env.example` tells an operator to generate it with that command and mount it at `/app/certs/oidc.pem`, with no mention of the mode; a file owned by the host user and unreadable by container uid 997 (the Google Wallet cert incident, same directory) then raises `PermissionError` while importing `revel.settings`, so web, celery, beat and telegram all crash-loop rather than the provider degrading to off. The PR body schedules "mount oidc.pem chmod 644 for uid 997" as an infra follow-up, so the trap is known; the self-hoster recipe in `.env.example` should say so, or the read should be guarded and reported through the existing `oauth.E001`-style system check.

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/src/oauth/management/commands/generate_oidc_signing_key.py#L30-L32

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/src/revel/settings/oauth.py#L22-L34

https://github.com/letsrevel/revel-backend/blob/30a5cbb3d79ffb46ee4bdd8df99e3866e019b0d2/.env.example#L136-L139

Below the reporting bar but worth a glance: the oauth controllers declare no error codes in `response=` (the `{detail, error}` 400 the consent page must branch on is absent from the generated client), and Grant/IDToken admins are registered but have no sidebar link.

